PR #5, opened 2024-01-10
Created NaadRssClient for fetching and parsing NAAD RSS feed

mhaswell-bcgov:
Added fetch() to fetch RSS XML, validate, and construct SimpleXMLElement Added getAlert() to get an alert from the RSS feed by id Added unit tests for fetch() and getAlert()
Added XML files for test data
Created get_rss.php for running NaadRssClient manually Renamed NaadClient to NaadSocketClient


src/get_rss.php
```diff
@@ -0,0 +1,9 @@
+<?php
+require_once 'vendor/autoload.php';
```
ShawnTurple, line 2: feels like wee need a bootstrap file, we should be only including autoloader once.

src/get_rss.php
```diff
@@ -0,0 +1,9 @@
...
+
+use Bcgov\NaadConnector\NaadRssClient;
+
+$rssClient = new NaadRssClient($argv[1]);
+if (1 === $rssClient->fetch()) {
+    print_r($rssClient->getAlert($argv[2]));
```
ShawnTurple, line 8: probably should create a logger function, as it will probably be required.

mhaswell-bcgov: Yeah I was going to suggest we create a ticket to figure out proper logging. Maybe we could try something like Monolog (https://github.com/Seldaek/monolog) unless we don't want to add any dependencies.

src/get_rss.php
mhaswell-bcgov: Not sure how we'd do this in this case since there are two entrypoints (one for the socket and one for the RSS). I don't think we'll end up needing the RSS entrypoint but I included it so that it would be possible to manually test.

ShawnTurple (approved):
This looks good as initial, however some considerations for next pr
* bootloader.php, which loads the classes if this is possible
* some sort of logging, keeping in mind what might be useful for different env ie dev/test/prod, and what gets outputted
* integration of rss feed to validate past alerts that the socket outputted

src/get_rss.php
ShawnTurple: i was kind of thinking this was the case, then technically there should be 2 composer files as these might live in different pods, i suspect time will sort this out.

src/get_rss.php
ShawnTurple: did you pick this one because of its ranking in most articles, or have you used it before, either way this seems logical

mhaswell-bcgov: Both. I've used it before and then did some searching and it seems to be one of the more commonly recommended ones.

PR #6, opened 2024-01-19
Added validateResponse() function to combine multi-part XML responses…

mhaswell-bcgov:
… and attempt to validate

Added unit tests for above

### DIFF
--- a/src/NaadSocketClient.php
+++ b/src/NaadSocketClient.php
@@ -41,6 +41,8 @@ class NaadSocketClient
      */
     protected int $port;
 
+    protected string $currentOutput = '';
+
     /**
      * Constructor for NaadClient.
      *
@@ -97,7 +99,9 @@ class NaadSocketClient
 
         $this->logger('Reading response:');
         while ( $out = socket_read($socket, self::$MAX_MESSAGE_SIZE) ) {
-            $this->handleEvent($out);
+            $previousUseInternalErrorsValue = libxml_use_internal_errors(true);
+            $this->handleResponse($out);
+            libxml_use_internal_errors($previousUseInternalErrorsValue);
         }
 
         $this->logger('Closing socket');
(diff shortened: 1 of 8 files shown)


ASpiteri-BCGov (approved):

src/NaadSocketClient.php
```diff
@@ -107,15 +116,53 @@ public function connect(): int
...
+     *
+     * @return bool
      */
-    protected function handleEvent( string $event )
+    protected function validateResponse( string $response ): bool
     {
-        $this->logger($event);
+        $this->currentOutput .= $response;
+        
+        $xml = simplexml_load_string($this->currentOutput);
```
robrien-bcgov, line 147: php seems to have a function for everything. It really makes coding backend stuff simple!

I bet in JS I'd have to import a library and learn its api, but with PHP it's just reading the documentation (and i guess reading the wordpress docs for frontend stuff)

src/NaadSocketClient.php
```diff
@@ -107,15 +116,53 @@ public function connect(): int
...
+
+        // Current output is not a valid XML document.
+        if (false === $xml) {
+            /**
+             * </alert> indicates the end of an alert XML document,
+             * clear current output for the next response.
+             */
+            if (str_ends_with(trim($this->currentOutput), '</alert>')) {
```
robrien-bcgov, line 155: but if $xml is false, doesn't that mean `simplexml_load_string()` received a malformed XML?

tests/NaadSocketClientTest.php
```diff
@@ -0,0 +1,87 @@
...
+                    [
+                        'location' => 'multipart/1.xml',
+                        'expected' => false
+                    ],
+                    [
+                        'location' => 'multipart/invalid/2.xml',
+                        'expected' => false
+                    ]
+                ]
+            ],
```
robrien-bcgov, line 67: i'll have to assume these test cases are all passing, and that's how you know your conditional works as it should in the NaadSocketClient

there is something to this test-driven development, but I've never felt I had time to try it out at work

robrien-bcgov (approved):

src/NaadSocketClient.php
mhaswell-bcgov: Yeah the logic is a little confusing. The problem I'm trying to solve is if we get some bad XML, we'll keep adding new socket responses onto the currentOutput and because there's already invalid XML in it, it will never pass validation unless we clear out currentOutput and let the responses start building a fresh XML document.
So this is saying if we have invalid XML but the response is trying to terminate with </alert>, we should clear the currentOutput so the next response can start fresh.

tests/NaadSocketClientTest.php
mhaswell-bcgov: We should probably add a workflow that runs the tests.

robrien-bcgov: that's a neat idea - it would demonstrate conclusively that the app passes tests, and even might help with PR reviews, because we could read the test and know it passes with the automated testing from github. good idea for some future pr!

src/NaadSocketClient.php
robrien-bcgov: ah ok that makesx sense

src/NaadSocketClient.php
```diff
@@ -97,7 +104,9 @@ public function connect(): int
 
         $this->logger('Reading response:');
         while ( $out = socket_read($socket, self::$MAX_MESSAGE_SIZE) ) {
-            $this->handleEvent($out);
+            $previousUseInternalErrorsValue = libxml_use_internal_errors(true);
```
ShawnTurple, line 107: this might be a great place for comment, to let other developer know what the purpose of the libxml_use_internal_errors, and why these lines of code are here

ShawnTurple: is this in order to handle multi-part response?

mhaswell-bcgov: This is just to be able to get libxml errors. Otherwise the xml library functions will return false but won't indicate what the actual cause of the failure was. I can add a comment to explain this in my next PR.

PR #7, opened 2024-01-23
Feature/descw 1870 identify heartbeats

mhaswell-bcgov:


src/NaadSocketClient.php
```diff
@@ -162,7 +178,22 @@ protected function validateResponse( string $response ): bool
...
+     * Determines whether a given SimpleXMLElement is a NAAD heartbeat message.
+     *
+     * @param SimpleXMLElement $xml XML from NAAD socket.
+     *
+     * @return boolean True if the XML is a heartbeat message, false otherwise.
+     */
+    protected function isHeartbeat( SimpleXMLElement $xml ): bool
+    {
+        $sender = $xml->xpath(
+            '/x:alert/x:sender[contains(text(),"NAADS-Heartbeat")]'
```
ShawnTurple, line 194: is there any mention of this distinction in the NAAD document (ie, to determine a heartbeat, check the sender text), or is this the most logical conclusion

src/NaadSocketClient.php
```diff
@@ -124,11 +131,19 @@ public function connect(): int
...
     {
-        if (!$this->validateResponse($response)) {
+        $xml = $this->validateResponse($response);
+        
+        if (!$xml) {
             return false;
         }
 
-        $this->logger($this->currentOutput);
+        $xml->registerXPathNamespace('x', 'urn:oasis:names:tc:emergency:cap:1.2');
```
ShawnTurple, line 140: is this a version at the end, if so what is the chance of it changing, if it does, would all the code fail?

ShawnTurple (approved):

src/NaadSocketClient.php
mhaswell-bcgov: This is an example of a heartbeat message: https://github.com/bcgov/naad-connector/blob/7668cd1962da459f09a67807ddb8130737f6abdc/tests/Socket/heartbeat.xml.
The only place that seems to indicate that it's a heartbeat is inside the `<sender>` element so I chose that. The NAAD documentation does mention using `<sender>` to check heartbeats but also mentions checking that `<status>` is "System" so maybe I should add that check too.

src/NaadSocketClient.php
mhaswell-bcgov: It's possible for the schema version to change which would cause the code to fail. The version is referring to this schema: https://docs.oasis-open.org/emergency/cap/v1.2/CAP-v1.2-os.html. It hasn't been changed since 2010.
My understanding is that the schema should be referenced explicitly as it's better for the code to fail completely than for the code to work with the wrong schema which could cause the code to do the wrong thing. For example if they changed the alert schema in such a way that the plugin filters no longer worked as expected it would be better for no events to be created than for an event to be created for every alert.

tests/Socket/heartbeat.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<alert
+	xmlns="urn:oasis:names:tc:emergency:cap:1.2"
```
ShawnTurple, line 3: this version looks like it belongs to the `urn:oasis:names:tc:emergency:cap` namespace, if this isn't true than i  don't see a problem

ShawnTurple (approved):

PR #8, opened 2024-01-24
Created DestinationClient to handle making requests to destination

mhaswell-bcgov:
Updated NaadSocketClient to make requests through DestinationClient Updated start.php to take CLI arguments for DestinationClient Updated Dockerfile and entrypoint.sh to use new CLI args Added test case for incorrect XML namespace to unit tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,5 +17,8 @@ RUN /usr/local/bin/composer dump-autoload
 
 ENV NAAD_NAME=NAAD-1
 ENV NAAD_URL=streaming1.naad-adna.pelmorex.com
+ENV DESTINATION_URL=https://localhost/eibc/wp-json/naad/v1/alert
+ENV DESTINATION_USER=naadbot
+ENV DESTINATION_PASSWORD='AAAA AAAA AAAA AAAA'
 
 ENTRYPOINT ["/home/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN /usr/local/bin/composer dump-autoload
 
 ENV NAAD_NAME=NAAD-1
 ENV NAAD_URL=streaming1.naad-adna.pelmorex.com
-ENV DESTINATION_URL=https://localhost/eibc/wp-json/naad/v1/alert
+ENV DESTINATION_URL=https://localhost/wp-json/naad/v1/alert
 ENV DESTINATION_USER=naadbot
 ENV DESTINATION_PASSWORD='AAAA AAAA AAAA AAAA'
 
--- a/src/DestinationClient.php
+++ b/src/DestinationClient.php
@@ -1,0 +1,94 @@
+<?php
+namespace Bcgov\NaadConnector;
+
+/**
+ * DestinationClient class makes requests to a destination url.
+ *
+ * @category Client
+ * @package  NaadConnector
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -17,5 +17,8 @@ RUN /usr/local/bin/composer dump-autoload
 
 ENV NAAD_NAME=NAAD-1
 ENV NAAD_URL=streaming1.naad-adna.pelmorex.com
+ENV DESTINATION_URL=https://localhost/eibc/wp-json/naad/v1/alert
```
ShawnTurple, line 20: i would try to refrain from using specifics here, even though it means you won't have to create an env for this.
```suggestion
ENV DESTINATION_URL=https://localhost/wp-json/naad/v1/alert
```

src/DestinationClient.php
```diff
@@ -0,0 +1,94 @@
...
+     * Sends an alert request to the destination.
+     *
+     * @param string $xml Raw XML string from alert.
+     *
+     * @return string
+     */
+    public function sendRequest(string $xml): string
+    {
+        // Open curl connection.
+        $curl = curl_init();
```
ShawnTurple, line 66: Can you verify that what you are doing with curl, is similar how a wp_request would do it under the hood, we are in a world we are not normally in, `a blank slate`

src/DestinationClient.php
```diff
@@ -0,0 +1,94 @@
...
+            sprintf('%s:%s', $this->username, $this->applicationPassword)
+        );
+    
+        // Set destination url.
+        curl_setopt($curl, CURLOPT_URL, $this->url);
+        curl_setopt($curl, CURLOPT_RETURNTRANSFER, 1);
+    
+        // POST to destination.
+        $result = curl_exec($curl);
+        print_r(curl_error($curl));
```
ShawnTurple, line 87: did you mean to leave this in, and if so maybe we should be using the logger

src/start.php
```diff
@@ -1,7 +1,9 @@
 <?php
 require_once 'vendor/autoload.php';
 
+use Bcgov\NaadConnector\DestinationClient;
 use Bcgov\NaadConnector\NaadSocketClient;
 
-$connector = new NaadSocketClient($argv[1], $argv[2]);
+$destinationClient = new DestinationClient($argv[3], $argv[4], $argv[5]);
```
ShawnTurple, line 7: not a fan of this, i would try to parse out these, because what if there wasn't argv[5], although i suspect you are controlling this as well at worse maybe something like `$argv[3]??''` <- i think this is the php context

ShawnTurple (approved):

src/DestinationClient.php
mhaswell-bcgov: I was able to test it with the route we created on the EIBC plugin and the request is working so I think in terms of compatibility with WordPress it's fine. We could use a library for this instead of writing a raw curl call but I think it's ok for now.

src/start.php
mhaswell-bcgov: Yeah it doesn't look pretty but I'm not sure we want the app to even run if all of the arguments aren't provided because it won't do what it's intended to do. I think it's better that the app just fails to start instead of running silently with arguments we know won't allow it to work as expected.

ShawnTurple (approved):

PR #9, opened 2024-02-27
Feature/descw 1868 deployments

ShawnTurple:


deployments/kustomize/base/app.yaml
```diff
@@ -8,24 +8,23 @@ metadata:
...
-          image: image-registry.openshift-image-registry.svc:5000/e9b785-tools/naad-app:latest
+          image: bcgovgdx/naad-app
           imagePullPolicy: Always
-          env:
-            - name: NAAD_NAME
-              value: NAAD-1
-            - name: NAAD_URL
-              value: "streaming1.naad-adna.pelmorex.com"
+          envFrom:
+          - configMapRef:
```
robrien-bcgov, line 20: a cleaner way

Dockerfile
```diff
@@ -12,6 +13,7 @@ RUN chmod +x /home/entrypoint.sh
 WORKDIR /var/www/html/
 
 COPY --from=composer:latest /usr/bin/composer /usr/local/bin/composer
+USER 1001
```
kshapka-bcgov, line 16: Does this solve some sort of permissions problem?

robrien-bcgov (approved):

kshapka-bcgov (approved):

Dockerfile
ShawnTurple: no, but it removes the root user, and still works, so its probably a win,

deployments/kustomize/base/app.yaml
ShawnTurple: if the keys in the configmap match exactly, then this is a nice way of doing it.

PR #10, opened 2024-03-12
DESCW-1901 Implement Monolog logging framework

kshapka-bcgov:
Note that this PR does not include changes to NaadRssClient, although incorporating monolog there should be fairly simple.
Reviewers: I have placed questions for you in comments where I was uncertain of what the functionality should be.

### DIFF
--- a/src/CustomLogger.php
+++ b/src/CustomLogger.php
@@ -18,30 +18,25 @@ use Monolog\Processor\PsrLogMessageProcessor;
 class CustomLogger
 {
     /**
-     * The Monolog logger instance.
-     *
-     * @var Logger
-     */
-    protected $logger;
-
-    /**
-     * Logger constructor.
-     *
+     * Return a Monolog Logger.
+     * 
      * @param string $channelName The name of the logging channel.
      * @param string $level       The minimum logging level to record.
+     * 
+     * @return Logger
      */
-    public function __construct(
+    public static function getLogger(
         string $channelName = 'monolog',
         string $level = 'info'
     ) {
         // Set up a monolog channel.
-        $this->logger = new Logger($channelName);
+        $logger = new Logger($channelName);
 
         // Processes a record's message according to PSR-3 rules.
         $processor = new PsrLogMessageProcessor();
-        $this->logger->pushProcessor($processor);
+        $logger->pushProcessor($processor);
 
-        $logLevel = $this->_convertLogLevel($level);
+        $logLevel = self::_convertLogLevel($level);
 
         // Store records to stdout.
         $stream = new StreamHandler('php://stdout', $logLevel);
@@ -50,17 +45,9 @@ class CustomLogger
         // For example, JSON formatter: $stream->setFormatter(new JsonFormatter());
         // Don't forget to include `use Monolog\Formatter\JsonFormatter` if you do.
 
-        $this->logger->pushHandler($stream);
-    }
-
-    /**
-     * Return a Monolog Logger.
-     *
-     * @return Logger
-     */
-    public function getLogger()
-    {
-        return $this->logger;
+        $logger->pushHandler($stream);
+        
+        return $logger;
     }
 
     /**
@@ -70,7 +57,7 @@ class CustomLogger
      * 
      * @return int
      */
-    private function _convertLogLevel($loggingLevel)
+    private static function _convertLogLevel($loggingLevel)
     {
         $normalizedLevel = strtolower($loggingLevel);
 
--- a/src/CustomLogger.php
+++ b/src/CustomLogger.php
@@ -1,0 +1,90 @@
+<?php
+namespace Bcgov\NaadConnector;
+
+use Monolog\{Level, Logger};
+use Monolog\Handler\StreamHandler;
+use Monolog\Processor\PsrLogMessageProcessor;
+
+/**
+ * A class for setting up a custom Monolog logger instance
+ * with configurable channel name and logging level.
+ *
+ * @category Logging
+ * @package  NaadConnector
+ * @author   Kyle Shapka <Kyle.Shapka@gov.bc.ca>
+ * @license  https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link     https://alerts.pelmorex.com/
+ */
+class CustomLogger
+{
+    /**
+     * The Monolog logger instance.
+     *
+     * @var Logger
+     */
+    protected $logger;
+
+    /**
+     * Logger constructor.
+     *
+     * @param string $channelName The name of the logging channel.
+     * @param string $level       The minimum logging level to record.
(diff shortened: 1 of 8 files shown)


src/NaadSocketClient.php
```diff
@@ -99,40 +114,54 @@ public function __construct(
      */
     public function connect(): int
     {
+        // Processes a record's message according to PSR-3 rules.
+        $processor = new PsrLogMessageProcessor();
+        $this->logger->pushProcessor($processor);
+
+        // Store records to stdout.
+        $stream = new StreamHandler('php://stdout', Level::Info);
```
kshapka-bcgov, line 122: This is where we can set the minimum logging level: could possibly be set by environmental variable in the future.

src/NaadSocketClient.php
```diff
@@ -99,40 +114,54 @@ public function __construct(
...
-        $this->logger('Creating socket');
+        $this->logger->info('Creating socket');
         $socket = socket_create(AF_INET, SOCK_STREAM, SOL_TCP);
         if ($socket === false ) {
-            $this->logger(
-                sprintf(
-                    'socket_create() failed: reason: %s',
-                    socket_strerror(socket_last_error())
-                )
+            $this->logger->info(
```
kshapka-bcgov, line 129: Should this be set to an error level log?

src/NaadSocketClient.php
```diff
@@ -197,10 +231,10 @@ protected function validateResponse( string $response ): bool|SimpleXMLElement
              * clear current output for the next response.
              */
             if (str_ends_with(trim($this->currentOutput), '</alert>')) {
-                $this->logger('Invalid XML document received.');
+                $this->logger->alert('Invalid XML document received.');
                 $this->currentOutput = '';
             } else {
-                $this->logger('Invalid or partial XML document received.');
+                $this->logger->error('Invalid or partial XML document received.');
```
kshapka-bcgov, line 237: What log level should this be? Not clear from context.

src/NaadSocketClient.php
```diff
@@ -210,12 +244,13 @@ protected function validateResponse( string $response ): bool|SimpleXMLElement
         $namespaces = $xml->getNamespaces();
         $capNamespace = $namespaces[""];
         if (self::$XML_NAMESPACE !== $capNamespace) {
-            $this->logger(
-                sprintf(
-                    "Unexpected namespace '$capNamespace'. " .
-                    "Expecting namespace '%s'.",
-                    self::$XML_NAMESPACE
-                )
+            $this->logger->info(
```
kshapka-bcgov, line 247: Again, what log level should this be?

src/NaadSocketClient.php
kshapka-bcgov: Note also logging to stdout but could specify a path-to-log-file

ShawnTurple: This code will be required every time we need logger, and we will need the exact same configuration. 
Finding a way to make the code more DRY'r so when we have to use it in other locations, we don't need to do all the initialization.

ShawnTurple:
Logging levels can be a bit tricky sometimes to get right, here are some of my thoughts.
* imagine yourself going through logs, either on prod or test, make sure the messages have meaning, and that they are triggered where it would be helpful
* levels use warning and error as your most critical, because this is what you really need to know to ensure your application is working appropriately (maybe an exception would be access logs, to indicate a successful hit to the WordPress endpoint) 

## Resources
* https://sematext.com/blog/logging-levels/ 
* https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels

src/NaadSocketClient.php
ShawnTurple: see my main message, about logging in general, i don't believe this should be an error, the stackoverflow has a good explination.

ShawnTurple (dismissed):
see my comments regarding logging levels

src/NaadSocketClient.php
mhaswell-bcgov: Yeah this should be an error. Anywhere the execution stops because something went wrong like this I think could be an error.

src/NaadSocketClient.php
```diff
@@ -259,7 +270,7 @@ protected function logger( string $msg )
     protected function logXmlErrors()
     {
         foreach (libxml_get_errors() as $error) {
-            print_r($error->message);
+            $this->logger->error($error->message);
```
mhaswell-bcgov, line 273: Despite the name I think this should actually be info level. These XML errors happen a lot because the the XML we receive from the socket isn't always complete. When we try to build a complete XML from multiple messages, the XML will be invalid until it's complete so we get these "errors" but they're really just issues with the validity of the XML.

src/start.php
```diff
@@ -3,7 +3,23 @@
...
+
+// Processes a record's message according to PSR-3 rules.
+$processor = new PsrLogMessageProcessor();
+$logger->pushProcessor($processor);
+
+// Store records to stdout.
+$stream = new StreamHandler('php://stdout', Level::Info);
+$logger->pushHandler($stream);
+
+$connector = new NaadSocketClient($argv[1], $argv[2], $destinationClient, $logger);
```
mhaswell-bcgov, line 24: This code will have to be repeated for the RSS class so I think ideally it would live in its own class with a static method that just returns this $logger object. Then it can be reused in both places.

mhaswell-bcgov (dismissed):

src/CustomLogger.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Return a Monolog Logger.
+     *
+     * @return Logger
+     */
+    public function getLogger()
+    {
+        return $this->logger;
```
mhaswell-bcgov, line 63: I think since all we really want is the resulting logger we could move the code that's currently in the constructor here and make this function static. Because the way it is now we're instantiating the `CustomLogger` object, then calling `getLogger()` from it and the rest of the code is using that logger, which leaves the `CustomLogger` object we created with no real purpose other than spawning that logger.

src/NaadRssClient.php
```diff
@@ -37,6 +45,8 @@ class NaadRssClient
     public function __construct( string $url )
     {
         $this->address = $url;
+        $customLogger = new CustomLogger('NaadRssClient', 'info');
+        $this->logger = $customLogger->getLogger();
```
mhaswell-bcgov, line 49: So if we made the change I suggested above this would a single line: `$this->logger = CustomLogger::getLogger('NaadRssClient', 'info');`.

mhaswell-bcgov (approved):

src/CustomLogger.php
kshapka-bcgov: Yeah, makes sense.

ShawnTurple (approved):

PR #11, opened 2024-04-19
DESCW-2311 update cpu, and cpu limits

ShawnTurple:


mhaswell-bcgov (approved):
I assume we need a ticket to fix the underlying issue where the script stopped executing but the pod didn't restart?

ShawnTurple:
> I assume we need a ticket to fix the underlying issue where the script stopped executing but the pod didn't restart?

yes this is a bigger issue, and something that we will need to write code for to give a health status, and a readiness state.  Usually this is done via a http:// call, but this is not a server, so we will have to do something with the heart beats, and timing

PR #12, opened 2024-09-20
Created liveness.sh to determine liveness based on last time log file…

mhaswell-bcgov:
… was modified

Added logging to file to CustomLogger
Refactored CustomLogger and NaadSocketClient constructors to allow passing log file path Updated deployments/kustomize/base/app.yaml and kustomization.yaml to allow the local image to be used when run locally Updated README with instructions on how to run in Kubernetes locally

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,5 +22,6 @@ ENV NAAD_URL=streaming1.naad-adna.pelmorex.com
 ENV DESTINATION_URL=https://localhost/wp-json/naad/v1/alert
 ENV DESTINATION_USER=naadbot
 ENV DESTINATION_PASSWORD='AAAA AAAA AAAA AAAA'
+ENV LOG_FILE_PATH=/var/www/html/app.log
 
 ENTRYPOINT ["/home/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,6 @@ ENV NAAD_URL=streaming1.naad-adna.pelmorex.com
 ENV DESTINATION_URL=https://localhost/wp-json/naad/v1/alert
 ENV DESTINATION_USER=naadbot
 ENV DESTINATION_PASSWORD='AAAA AAAA AAAA AAAA'
-ENV LOG_FILE_PATH=/var/www/html/app.log
+ENV LOG_FILE_PATH=/var/log/naad-socket.log
 
 ENTRYPOINT ["/home/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ ENV NAAD_NAME=NAAD-1
 ENV NAAD_URL=streaming1.naad-adna.pelmorex.com
 ENV DESTINATION_URL=https://localhost/wp-json/naad/v1/alert
 ENV DESTINATION_USER=naadbot
-ENV DESTINATION_PASSWORD='AAAA AAAA AAAA AAAA'
-ENV LOG_FILE_PATH=/var/log/naad-socket.log
+ENV DESTINATION_PASSWORD="AAAA AAAA AAAA AAAA"
+ENV LOG_FILE_PATH="/var/www/html/naad-socket.log"
 
 ENTRYPOINT ["/home/entrypoint.sh"]
--- a/deployments/kustomize/base/app.yaml
+++ b/deployments/kustomize/base/app.yaml
@@ -15,7 +15,7 @@ spec:
       containers:
         - name: naad-app
           image: bcgovgdx/naad-app
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
           envFrom:
           - configMapRef:
               name: config
@@ -28,3 +28,9 @@ spec:
             limits:
               cpu: 500m
               memory: 1Gi
+          livenessProbe:
+            exec:
+              command:
+              - /var/www/html/liveness.sh
+            initialDelaySeconds: 60
+            periodSeconds: 60
--- a/deployments/kustomize/base/kustomization.yaml
+++ b/deployments/kustomize/base/kustomization.yaml
@@ -11,7 +11,7 @@ configMapGenerator:
   - name: config
     literals:
     - NAAD_NAME=NAAD-1
-    - NAAD_URL=streaming2.naad-adna.pelmorex.com
+    - NAAD_URL=streaming1.naad-adna.pelmorex.com
     - DESTINATION_URL=https://localhost/wp-json/naad/v1/alert
     - DESTINATION_USER=naadbot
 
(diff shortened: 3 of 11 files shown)


Dockerfile
```diff
@@ -22,5 +22,6 @@ ENV NAAD_URL=streaming1.naad-adna.pelmorex.com
 ENV DESTINATION_URL=https://localhost/wp-json/naad/v1/alert
 ENV DESTINATION_USER=naadbot
 ENV DESTINATION_PASSWORD='AAAA AAAA AAAA AAAA'
+ENV LOG_FILE_PATH=/var/www/html/app.log
```
ShawnTurple, line 25: so generally speaking logs are not in the same directory as html folder, they are in a non reachable folder to the public, i know that this technically won't be, but as a default i think we should follow linux best practices.  please consider this point, but no need to change this moment

deployments/kustomize/base/app.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       containers:
         - name: naad-app
           image: bcgovgdx/naad-app
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
```
ShawnTurple, line 18: did this fix the issue you were having with building?

entrypoint.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-/usr/local/bin/php /var/www/html/src/start.php $NAAD_NAME $NAAD_URL $DESTINATION_URL $DESTINATION_USER $DESTINATION_PASSWORD
\ No newline at end of file
+/usr/local/bin/php /var/www/html/src/start.php $NAAD_NAME $NAAD_URL $DESTINATION_URL $DESTINATION_USER $DESTINATION_PASSWORD $LOG_FILE_PATH
```
ShawnTurple, line 3: i suspect there is a better way of doing this, i wonder if we have an env file which gets consumed or are these env variables simply exposed to the php without doing anything.... one way is to have a conf.php, similar to wordpress where they define globals... maybe for future considerations.

liveness.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+LAST_MODIFIED=$(date -r $LOG_FILE_PATH +%s)
+CURRENT=$(date +%s)
+
+# If the log file hasn't been modified in the last 2 minutes, liveness fails.
+if [ $(($LAST_MODIFIED + 120)) -ge $CURRENT ]
```
ShawnTurple, line 7: are you sure you just can't equate this to a variable and echo it out, to provide the same results 
```suggestion
echo  $(($LAST_MODIFIED + 120)) -ge $CURRENT 
```
it might need a bit of bash formatting to work, but technically this is what this probably does anyway.

src/CustomLogger.php
```diff
@@ -27,7 +27,8 @@ class CustomLogger
      */
     public static function getLogger(
         string $channelName = 'monolog',
-        string $level = 'info'
+        string $level = 'info',
+        string $logFilePath = './app.log'
```
ShawnTurple, line 31: how come you are not using the ENV for the logfile here?

src/CustomLogger.php
```diff
@@ -62,16 +67,16 @@ private static function _convertLogLevel($loggingLevel)
...
         $logLevels = [
-            'emergency' => Logger::EMERGENCY,
-            'alert'     => Logger::ALERT,
-            'critical'  => Logger::CRITICAL,
-            'error'     => Logger::ERROR,
-            'warning'   => Logger::WARNING,
-            'notice'    => Logger::NOTICE,
-            'info'      => Logger::INFO,
-            'debug'     => Logger::DEBUG,
+            'emergency' => Level::Emergency,
```
ShawnTurple, line 70: i'm not sure the purpose of this, why not just at line 30 say `Level::info`, if this is the only place you are going to set it.

src/start.php
```diff
@@ -1,9 +1,12 @@
...
 require_once 'vendor/autoload.php';
 
+use Bcgov\NaadConnector\CustomLogger;
 use Bcgov\NaadConnector\DestinationClient;
 use Bcgov\NaadConnector\NaadSocketClient;
 
 $destinationClient = new DestinationClient($argv[3], $argv[4], $argv[5]);
-$connector = new NaadSocketClient($argv[1], $argv[2], $destinationClient);
+
+$socketLogger = CustomLogger::getLogger('NaadSocketClient', 'info', $argv[6]);
```
ShawnTurple, line 10: not sure why we would want to set the logging level, different depending on the file, unless i'm misunderstanding something.

ShawnTurple (dismissed):
please consider all my comments, before merging

deployments/kustomize/base/kustomization.yaml
```diff
@@ -21,3 +21,6 @@ secretGenerator:
     literals:
     - DESTINATION_PASSWORD='AAAA AAAA AAAA AAAA'
 
+images:
+  - name: bcgovgdx/naad-app
+    newName: naad-app
```
kshapka-bcgov, line 26: I didn't know this was a thing, cool

liveness.sh
mhaswell-bcgov: For some reason the liveness probe wants a return value of 0 when the liveness check is successful and a return value of anything else when it's not so the logic would look kind of weird I think without the if.

src/CustomLogger.php
mhaswell-bcgov: I changed these because Monolog deprecated accessing the log levels enum directly through the Logger class. This function for setting log levels already existed but I wanted to get rid of the squigglies.

deployments/kustomize/base/app.yaml
mhaswell-bcgov: This was part of it. The other part was that the app requires PHP 8+ and I was on 7.4.

entrypoint.sh
mhaswell-bcgov: I agree, passing the variables in as arguments is going to get out of hand quickly. I'll create a ticket to look into a better way of using the env variables. Docker also complains about how we're using the $DESTINATION_PASSWORD when we build the image: ` - SecretsUsedInArgOrEnv: Do not use ARG or ENV instructions for sensitive data (ENV "DESTINATION_PASSWORD")`.

Dockerfile
mhaswell-bcgov: Changed this to default to `/var/log/naad-socket.log`.

ShawnTurple (approved):
did you run a linter the second time?

mhaswell-bcgov:
> did you run a linter the second time?

@ShawnTurple Yeah I forgot to run phpcs before then. I'll also write a ticket to add a GH workflow check to make sure phpcs is passing.

ShawnTurple (approved):

PR #13, opened 2024-09-23
Feature/descw 1902 database deployment

mhaswell-bcgov:


deployments/kustomize/base/app/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
...
+configMapGenerator:
+  - name: config
+    literals:
+    - NAAD_NAME=NAAD-1
+    - NAAD_URL=streaming1.naad-adna.pelmorex.com
+    - DESTINATION_URL=https://localhost/wp-json/naad/v1/alert
+    - DESTINATION_USER=naadbot
+
+secretGenerator:
+  - name: naad-secrets
```
ShawnTurple, line 19: does the word secrets need to be in the secret, as the type is a secret, so i suspect is redundant

deployments/kustomize/base/database/deployment.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: mariadb-deployment
```
ShawnTurple, line 4: is this a best practice to put deployment at the end of considering it is a deployment... If you give me a clear indication why we need to do this, otherwise i believe we could remove deployment.

deployments/kustomize/base/database/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
...
+configMapGenerator:
+  - name: database-config
+    literals:
+    - MARIADB_ALLOW_EMPTY_ROOT_PASSWORD="0" # If it is 1 and root_password is set, root_password takes precedence.
+
+secretGenerator:
+  - name: database-secrets
+    type: Opaque
+    literals:
+    - MARIADB_ROOT_PASSWORD="$z8ez=^)|YPPD]9NsJ*!.IWOpD.lgK(yH\k9J777g8(i3OJW["
```
ShawnTurple, line 22: since this password would only be used for local, is there a way to update it, so it doesn't look like a production password or forgot to change it.

deployments/kustomize/base/kustomization.yaml
```diff
@@ -1,22 +1,8 @@
...
-    - DESTINATION_USER=naadbot
-
-secretGenerator:
-  - name: secrets
-    type: Opaque
-    literals:
-    - DESTINATION_PASSWORD='AAAA AAAA AAAA AAAA'
\ No newline at end of file
+# Overrides the namespace
+namespace: default
```
ShawnTurple, line 8: why did you remove the commonLabels?

src/database_connection_check.php
```diff
@@ -0,0 +1,33 @@
...
+require_once 'vendor/autoload.php';
+
+use Doctrine\DBAL\DriverManager;
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\ORMSetup;
+
+// Create a simple "default" Doctrine ORM configuration.
+$config = ORMSetup::createAttributeMetadataConfiguration(
+    paths: [__DIR__ . '/src'],
+    isDevMode: true,
```
ShawnTurple, line 11: i'm assuming later you are going to have an env to change this?

ShawnTurple (dismissed):

ShawnTurple (dismissed):

deployments/kustomize/base/database/deployment.yaml
mhaswell-bcgov: Removed all of these names that reference their own resource type that I think I can.

deployments/kustomize/base/database/kustomization.yaml
mhaswell-bcgov: Changed to `rootpassword` so there's no confusion.

deployments/kustomize/base/kustomization.yaml
mhaswell-bcgov: Added them back although I don't fully understand what they're doing.

src/database_connection_check.php
mhaswell-bcgov: Yeah that would make sense. I'll have to look into what isDevMode actually does. I just took this from the Doctrine docs.

ShawnTurple (approved):

PR #15, opened 2024-10-04
Feature/descw 1905 socket database

mhaswell-bcgov:


data/doctrine/migrations/Version20241004171841.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Increase size of alerts.body column';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('
+            ALTER TABLE alerts MODIFY body MEDIUMTEXT NOT NULL;
```
ShawnTurple, line 23: it looks like mediumtext is the largest text type, i wasn't aware that they were broken down into various size limits

src/Database.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace Bcgov\NaadConnector;
+
+use Doctrine\DBAL\Connection;
```
ShawnTurple, line 5: although this is fine, i have seen a shorter way to include use statements https://www.php.net/manual/en/language.namespaces.importing.php
image

src/Database.php
```diff
@@ -0,0 +1,49 @@
...
+        // Create a simple "default" Doctrine ORM configuration.
+        $config = ORMSetup::createAttributeMetadataConfiguration(
+            paths: [__DIR__ . '/src'],
+            isDevMode: true,
+        );
+
+        // Configuring the database connection.
+        $connection = DriverManager::getConnection(
+            [
+            // TODO: Create non-root user and replace with env variable.
```
ShawnTurple, line 38: do todo's get caught by phpcs if they are not in the header?

src/Entity/Alert.php
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * Sets the send attempted date and time.
+     *
+     * @param \DateTime|null $sendAttempted The last time an alert was
+     *                                      attempted to send to its destination.
+     *
+     * @return void
+     */
+    public function setSendAttempted(?\DateTime $sendAttempted): void
```
ShawnTurple, line 126: unfortunately i suspect there is no way of giving a int value to say try 3 x if not successful than give up, a date is kind of like a boolean, where it either worked or it didn't, will you have logic that will keep trying to send until ~

src/Entity/Alert.php
```diff
@@ -0,0 +1,197 @@
...
+     * Creates an alert from a SimpleXMLElement.
+     *
+     * @param SimpleXMLElement $xml XML to create alert from.
+     *
+     * @return Alert
+     */
+    public static function fromXml(SimpleXMLElement $xml): Alert
+    {
+        $alert = new Alert();
+        $id = $xml->xpath(
```
ShawnTurple, line 187: there doesn't seem to be any logic for catching the other two possibilities of null or false for xpath  https://www.php.net/manual/en/simplexmlelement.xpath.php

src/Repository/AlertRepository.php
```diff
@@ -0,0 +1,55 @@
...
+     * Constructs an AlertRepository object.
+     *
+     * @param ManagerRegistry $registry ManagerRegistry object to construct with.
+     */
+    public function __construct(ManagerRegistry $registry)
+    {
+        parent::__construct($registry, Alert::class);
+    }
+
+    //    /**
```
ShawnTurple, line 31: is there a reason we are leaving all the commented code, if so, we should indicate why, otherwise please remove

tests/NaadSocketClientTest.php
```diff
@@ -21,7 +23,7 @@ public function testValidateResponse(array $xmlResponses) {
         $class = new ReflectionClass('Bcgov\NaadConnector\NaadSocketClient');
         $method = $class->getMethod('validateResponse');
         $logger = CustomLogger::getLogger();
-        $client = new NaadSocketClient('test-naad', 'testing.url', new DestinationClient('testing.url', 'user', 'pass'), $logger);
+        $client = new NaadSocketClient('test-naad', 'testing.url', new DestinationClient('testing.url', 'user', 'pass'), $logger, new Database());
```
ShawnTurple, line 26: this makes it more brief, but probably at the expense of readability.

ShawnTurple: are you using the actual database for tests?

tests/NaadSocketClientTest.php
```diff
@@ -131,5 +133,32 @@ public function testIsHeartbeat() {
         $result = $method->invokeArgs($client, [$alert]);
         $this->assertFalse($result);
     }
+
+    /**
+     * Undocumented function
```
ShawnTurple, line 138: adding comments to the function would be helpful

tests/NaadSocketClientTest.php
```diff
@@ -131,5 +133,32 @@ public function testIsHeartbeat() {
...
+     * @return void
+     */
+    public function testHandleResponse() {
+        $this->markTestSkipped('TODO: figure out how to mock database for testing.');
```
ShawnTurple, line 142: so does this just make the test work, meaning it doesn't fail, if so this is handy.

ShawnTurple (approved):
is there a github action that runs phpcs, and unit tests, if so, i would try to ensure this becomes a priority.

src/Database.php
mhaswell-bcgov: Not with the PHPCS ruleset we're using.

src/Entity/Alert.php
mhaswell-bcgov: My thinking is that the `success` column is a boolean that indicates whether the alert was successfully sent to the WP site. If `success` is true then we don't need to attempt to send it again. If `success` is false then we try to send it and if it fails, we increment the `failures` column to keep track of how many times it has failed.

src/Entity/Alert.php
mhaswell-bcgov: Yeah this is pretty fragile right now as I was just trying to get something inserting. I'll create a ticket to improve this function.

src/Repository/AlertRepository.php
mhaswell-bcgov: This file was generated from a Doctrine command and it came with these commented examples. I think we'll need to implement these in order to retrieve alerts from the database so I think it makes sense to leave the examples here until then.

tests/NaadSocketClientTest.php
mhaswell-bcgov: None of the existing tests interact with the database except the one that is being skipped until we figure this out. It would be using the real database if the tests were run in an environment that has a database running and the tests actually used it.

tests/NaadSocketClientTest.php
mhaswell-bcgov: It makes it so the test doesn't fail but the output of running phpunit looks like this:
`OK, but some tests were skipped!
Tests: 14, Assertions: 23, Skipped: 1.`

src/Entity/Alert.php
ShawnTurple: so do we have a max number of failures, or do we do it until we run out of computing power

src/Database.php
ShawnTurple: i think if they are in the header of the docbloc they do.

src/Entity/Alert.php
mhaswell-bcgov: There will have to be some maximum number of failures but I'm not sure what should happen then. If the requests aren't going through to the site then someone needs to intervene and fix it.

tests/NaadSocketClientTest.php
mhaswell-bcgov: Added a doc comment here.

tests/NaadSocketClientTest.php
mhaswell-bcgov: I ended up overriding the database env variables in phpunit.xml so that it shouldn't be possible to connect to a real database during tests.

ShawnTurple (approved):

PR #16, opened 2024-10-11
Feature/descw 2672 naad connector docker compose

robrien-bcgov:
hey @mhaswell-bcgov take a look at this and let me know what you think.
image

it's looking functional, and it should run on your machine when you type `docker compose up` in the root of the project.

I expect the myphpadmin to work as well, and it should show the migrations after about 5 seconds (a health check was added to the mariadb service)

to run phpMyAdmin, just go to http://0.0.0.0:8081

(kill your wordpress containers first with `docker compose down` in wordpress/dev)


docker-compose.yaml
```diff
@@ -0,0 +1,76 @@
...
+      MARIADB_ROOT_PASSWORD: ${MARIADB_ROOT_PASSWORD} # Loaded from .env
+      MARIADB_SERVICE_HOST: mariadb
+      MARIADB_SERVICE_PORT: 3306
+      NAAD_NAME: NAAD-1
+      NAAD_URL: streaming1.naad-adna.pelmorex.com
+    depends_on:
+      database-migrate:
+        condition: service_completed_successfully
+    ports:
+      - "8080:80"
```
ShawnTurple, line 51: we could add a env variable here as well 
```suggestion
      - "${NAAD_PORT:-8080}:80"
```

docker-compose.yaml
```diff
@@ -0,0 +1,76 @@
+services:
+  mariadb:
+    image: mariadb:11.5.2
```
ShawnTurple, line 3: if you add the full latest to the mariadb, this has its pros and cons, on one hand its a specific version, but on the other, every time version changes we have to change it, maybe this might be better
```suggestion
    image: mariadb:11
```

docker-compose.yaml
```diff
@@ -0,0 +1,76 @@
...
+    environment:
+      MARIADB_ALLOW_EMPTY_ROOT_PASSWORD: "0"
```
ShawnTurple, line 5: i believe by default it is set to false / 0, so why would we include it, there is no harm, but just feels like it might be better to not include it.

docker-compose.yaml
```diff
@@ -0,0 +1,76 @@
...
+      MARIADB_DATABASE: naad_connector
+      MARIADB_ROOT_PASSWORD: ${MARIADB_ROOT_PASSWORD}
+    ports:
+      - "3306:3306"
```
ShawnTurple, line 9: does this cause a conflict with the wordpress development, as we already expose 3306, however i could be wrong

docker-compose.yaml
```diff
@@ -0,0 +1,76 @@
...
+    networks:
+      - naad_deploy
+    ports:
+      - "8081:80"
+
+networks:
+  naad_deploy:
+
+volumes:
+  database-volume:
```
ShawnTurple, line 76: might want to be a bit more descriptive with this volume name, if you were to look at all your docker volumes, it might not be obvious this is for naad
```suggestion
  naad-database:
```

docker-compose.yaml
```diff
@@ -0,0 +1,76 @@
...
+        condition: service_healthy
+    entrypoint: ["vendor/bin/doctrine-migrations", "-vvv", "-n", "migrate"]
+    networks:
+      - naad_deploy
+
+  # wait for database migration to complete
+  naad-app:
+    image: bcgovgdx/naad-app
+    environment:
+      DESTINATION_URL: https://0.0.0.0/wp-json/naad/v1/alert
```
ShawnTurple, line 39: this should have an env variable 
```suggestion
      DESTINATION_URL: ${DESTINATION_URL:-https://0.0.0.0/wp-json/naad/v1/alert}
```

docker-compose.yaml
```diff
@@ -0,0 +1,76 @@
...
+    depends_on:
+      mariadb:
+        condition: service_healthy
+    entrypoint: ["vendor/bin/doctrine-migrations", "-vvv", "-n", "migrate"]
+    networks:
+      - naad_deploy
+
+  # wait for database migration to complete
+  naad-app:
+    image: bcgovgdx/naad-app
```
ShawnTurple, line 37: you shouldn't rely on this image, if it requires an image that we create, than it should use a build similar to this https://github.com/bcgov/wordpress/blob/main/dev/docker-compose.yaml#L13. however this might be another ticket 
```suggestion
    build: 
        context: ./
```
this might work, though without testing it, there might be some issues.

docker-compose.yaml
```diff
@@ -0,0 +1,76 @@
...
+    healthcheck:
+      test: ["CMD-SHELL", "mariadb-admin -u root -p${MARIADB_ROOT_PASSWORD} ping -h 127.0.0.1"]
+      interval: 10s
+      timeout: 5s
+      retries: 3
+    restart: always
+
+  # wait for mariadb to start up
+  database-migrate:
+    image: bcgovgdx/naad-app
```
ShawnTurple, line 23: this would use the same build / context as the app.

composer.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "bcgov/naad-connector",
-    "version": "1.0.0",
+    "version": "1.0.1",
```
ShawnTurple, line 3: don't need to change the version, leave it at 1.0.0

docker-compose.yaml
robrien-bcgov: yes it does, for a few seconds. then it exits.

I like to think that's efficient use of resources, but if you have a detailed plan of how to do this better, please write it down here as a reply and I'll try it out immediately.

docker-compose.yaml
robrien-bcgov: I'll try it and see what happens

robrien-bcgov: It doesn't work. What would you suggest I do instead, specifically?


image

ShawnTurple: we can leave this for now, the problem with this, is the code will always be only as the latest code pushed to dockerhub. --- i will create a ticket.

ShawnTurple (approved):

PR #17, opened 2024-10-21
Added feature to recover missed alerts through heartbeat messages

mhaswell-bcgov:
- NaadSocketClient uses heartbeat message's `references` property to retrieve last 10 alert ids
  - Checks if any of those ids are missing from the database
  - Fetches any missing alerts from NAAD's alert repository
  - Inserts them into the database Updated Database class to perform database queries so database interactions are contained in one class

### DIFF
--- a/src/Database.php
+++ b/src/Database.php
@@ -64,6 +64,8 @@ class Database
      * @param Alert $alert The alert to persist.
      *
      * @return string
+     * 
+     * @see https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/working-with-objects.html#persisting-entities
      */
     protected function persistAlert(
         Alert $alert
--- a/src/Database.php
+++ b/src/Database.php
@@ -2,13 +2,14 @@
 
 namespace Bcgov\NaadConnector;
 
-use Doctrine\DBAL\Connection;
+use Bcgov\NaadConnector\Entity\Alert;
 use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\ORMSetup;
 
 /**
- * Database class mainly used to handle connecting to the database.
+ * Database class for interacting with the database.
  * 
  * @category Database
  * @package  NaadConnector
@@ -19,12 +20,22 @@ use Doctrine\ORM\ORMSetup;
 class Database
 {
 
+    protected EntityManager $entityManager;
+
+    /**
+     * Constructor for Database class.
+     */
+    public function __construct()
+    {
+        $this->entityManager = $this->getEntityManager();
+    }
+
     /**
      * Gets an instance of the Doctrine EntityManager.
      *
      * @return EntityManager
      */
-    static function getEntityManager(): EntityManager
+    protected function getEntityManager(): EntityManager
     {
         // Create a simple "default" Doctrine ORM configuration.
         $config = ORMSetup::createAttributeMetadataConfiguration(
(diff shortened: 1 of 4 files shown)


src/Database.php
```diff
@@ -46,4 +57,50 @@ static function getEntityManager(): EntityManager
         );
         return new EntityManager($connection, $config);
     }
+
+    /**
+     * Persists an alert to the database.
```
ShawnTurple, line 62: if you knew nothing about what this does, would you be content with this docbloc and its description

src/NaadSocketClient.php
```diff
@@ -301,6 +314,100 @@ protected function isHeartbeat( SimpleXMLElement $xml ): bool
...
+     * @return array An array containing the heartbeat references for any alerts
+     *               that are not already in the database.
+     */
+    protected function findMissedAlerts(SimpleXMLElement $xml): array
+    {
+        $rawReferences = explode(' ', $xml->references);
+        $references = [];
+
+        // Separate the references value into sender, id, and sent parts.
+        foreach ($rawReferences as $reference) {
```
ShawnTurple, line 331:
```suggestion
        foreach ($rawReferences as $rawReference) {
```
this might be a better name, as this might fit better with the foreach ... it seems to close to $references = []

src/NaadSocketClient.php
```diff
@@ -301,6 +314,100 @@ protected function isHeartbeat( SimpleXMLElement $xml ): bool
...
+            $referenceParts = explode(',', $reference);
+            $references[] = [
+                'sender' => $referenceParts[0],
```
ShawnTurple, line 334: can you be 100% sure that there is a $referenceParts with 3 elements, and these elements are always as you indicated.

src/NaadSocketClient.php
```diff
@@ -301,6 +314,100 @@ protected function isHeartbeat( SimpleXMLElement $xml ): bool
...
+                'id' => $referenceParts[1],
+                'sent' => $referenceParts[2],
+            ];
+        }
+
+        // Remove any reference ids that already exist in the database.
+        $existingAlerts = $this->database->getAlertsById(
```
ShawnTurple, line 341: so every heartbeat you are pocking the database 10x + times to verify it was successful, would it be better to get the last 10 items from the database and check with the heartbeat, it might not be an issue this way, just seems weird to have to probe a database for every check... unless i'm reading this wrong.

src/NaadSocketClient.php
mhaswell-bcgov: This is the way they're structured in the heartbeat messages:
`<references>cap-pac@canada.ca,urn:oid:2.49.0.1.124.3391214003.2024,2024-01-17T16:05:28-00:00 cap-pac@canada.ca,urn:oid:2.49.0.1.124.3512422613.2024,2024-01-17T16:08:14-00:00 cap-pac@canada.ca,urn:oid:2.49.0.1.124.3964945571.2024,2024-01-17T16:09:58-00:00 ...</references>`
but I'm not sure if we can say 100% they will always be formatted correctly. I don't know how else to do this that would be able to handle that changing though.

src/Database.php
mhaswell-bcgov: Added a link to the Doctrine docs on persisting objects.

src/NaadSocketClient.php
mhaswell-bcgov: Made this change.

src/NaadSocketClient.php
ShawnTurple:
```suggestion
                'sender' => $referenceParts[0] ?? false,
```
might be a way... though check my syntax to make sure this is right... and then we could check if not false to continue on.

ShawnTurple (approved):

src/NaadSocketClient.php
mhaswell-bcgov: How about this:
```
            if (count($referenceParts) < 3) {
                $this->logger->critical(
                    'Heartbeat reference part count too low. Expected 3, but got {count}.',
                    ['count' => count($referenceParts)]
                );
                exit(1);
            }
```
It'll cause the pod to die if we don't get the 3 reference parts that we expect. If we continue on when we get something unexpected here then we won't be able to retrieve missed alerts.

src/NaadSocketClient.php
```diff
@@ -301,6 +316,100 @@ protected function isHeartbeat( SimpleXMLElement $xml ): bool
...
+     * @param array $reference Heartbeat references array parts (sender, id, sent).
+     *
+     * @return boolean|SimpleXMLElement
+     */
+    protected function fetchAlertFromRepository(
+        array $reference
+    ): bool|SimpleXMLElement {
+        $url = sprintf(
+            'http://%s/%s/%sI%s.xml',
+            'capcp1.naad-adna.pelmorex.com',
```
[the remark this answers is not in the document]
ShawnTurple: but if there is no way of changing what is the point, either make it use an env or don't

src/NaadSocketClient.php
ShawnTurple: if this make sense.  i'm just making sure if this condition happens, it does more than just puke

PR #18, opened 2024-10-23
DESCW-2623 DESCW-2658 Add github workflows for PHPCS and PHPUNIT

kshapka-bcgov:


composer.json
```diff
@@ -15,29 +21,36 @@
         }
     ],
     "require-dev": {
-        "phpunit/phpunit": "^10",
-        "squizlabs/php_codesniffer": "^3.9"
+        "bcgov/wordpress-utils": "^2.1.3"
```
mhaswell-bcgov, line 24: I'd rather not use wordpress-utils here. I'm trying to mostly keep it to whatever the most basic, default PHPCS rules are since this code is PHP but isn't really related to WordPress.

.github/workflows/php-lint-test.yml
```diff
@@ -0,0 +1,35 @@
...
+        id: composer-cache
+        uses: actions/cache@v3
+        with:
+          path: vendor
+          key: ${{ runner.os }}-php-${{ hashFiles('**/composer.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-php-
+
+      - name: Install dependencies
+        run: composer update --prefer-dist --no-progress
```
mhaswell-bcgov, line 29: Does this still work if it's changed to `composer install`? From my reading of this it's going to be downloading whatever the latest versions of dependencies are from composer.json instead of the versions in the lock file.

.github/workflows/php-lint-test.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Validate composer.json and composer.lock
+        run: composer validate
+
+      - name: Cache Composer packages
+        id: composer-cache
+        uses: actions/cache@v3
+        with:
+          path: vendor
+          key: ${{ runner.os }}-php-${{ hashFiles('**/composer.lock') }}
```
mhaswell-bcgov, line 24: I haven't really used caching before. This caches the vendor/ folder based on a hash of the lock file? That makes sense to me but if this step happens before the dependencies are installed, does the vendor/ folder have anything in it yet?

mhaswell-bcgov (approved):

.github/workflows/php-lint-test.yml
kshapka-bcgov: Install failed when I first used it, although this might be due to the php requirement. I can try changing it back.

.github/workflows/php-lint-test.yml
kshapka-bcgov: It's funny you say that :smile:
https://github.com/bcgov/bcgov-wordpress-block-theme/blame/development/.github/workflows/php-lint-test.yml#L20

mhaswell-bcgov: My own code back to haunt me. I must have copied this from somewhere because I have no recollection of writing this. I guess we'll be able to see from the workflow output if it works or not.

.github/workflows/php-lint-test.yml
mhaswell-bcgov: Looks like the default version of PHP the runner uses is 8.1.2 but the version the composer.lock was generated on was my PHP 8.3.11, so some of the dependencies are requiring a later version of PHP that the workflow runner's environment doesn't satisfy. We might need to try adding a step like this:
`
    - name: Setup PHP
      uses: shivammathur/setup-php@v2
      with:
        php-version: 8.3
`

.github/workflows/php-lint-test.yml
kshapka-bcgov: It seems like it 'worked' in the sense that it didn't prevent the workflow from completing, but not in the sense that it accomplished anything. Removed.

mhaswell-bcgov (approved):
Nice, looks good to me.

PR #19, opened 2024-10-25
Adjusted socket message logging:

mhaswell-bcgov:
- When an invalid but complete alert is received, log at error level
  - When an invalid but incomplete alert is received, log at debug level (this is expected to occur when receiving larger alerts)
  - Specific XML validity errors caused by above are logged at debug level


robrien-bcgov (approved):
LGTM

PR #20, opened 2024-10-31
Created deployments/kustomize/base/phpmyadmin files:

mhaswell-bcgov:
- deployment.yaml for the PHPMyAdmin pod
  - kustomization.yaml for configurations
  - services.yaml to expose PHPMyAdmin pod to the host machine

Updated README to explain how to access PHPMyAdmin


README.md
```diff
@@ -30,6 +30,8 @@ docker build -t bcgovgdx/naad-app .
 kubectl apply -k deployments/kustomize/base/
 ```
 
+PHPMyAdmin will then be accessible at http://localhost:31008. You may need to use Firefox or Safari to access this as Chrome may block this address due to it not using https.
```
robrien-bcgov, line 33:
```suggestion
PHPMyAdmin will then be accessible at http://0.0.0.0:31008. You may need to use Firefox or Safari to access this as Chrome may block this address due to it not using https.
```

robrien-bcgov: using that address seemed to work for me on chrome

robrien-bcgov (approved):
looks legit

deployments/kustomize/base/phpmyadmin/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
ShawnTurple, line 1: i don't think this should be in the base, it should only live in the dev deployment

README.md
mhaswell-bcgov: This works for me too. Good fix!

robrien-bcgov (dismissed):

deployments/kustomize/base/phpmyadmin/kustomization.yaml
mhaswell-bcgov: Not sure how to go about doing this. I see there's a dev directory in the tenant repo, should all of the code here be moved there? If so, we would have to deploy to local Kubernetes through the tenant repo instead of through this one?

deployments/kustomize/overlays/local/phpmyadmin/deployment.yaml
```diff
@@ -0,0 +1,30 @@
...
+    matchLabels:
+      app: phpmyadmin
+  template:
+    metadata:
+      labels:
+        app: phpmyadmin
+    spec:
+      containers:
+        - name: phpmyadmin
+          image: phpmyadmin/phpmyadmin
```
ShawnTurple, line 19:
```suggestion
          image: phpmyadmin/phpmyadmin:5.2
```
i think it is always good to put a tagged version, otherwise somwhere down the line this could stop working, because of a major release... it also indicates at least a ball park version

deployments/kustomize/overlays/local/phpmyadmin/deployment.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: phpmyadmin
+  labels:
+    app.kubernetes.io/version:	11.5.2
```
ShawnTurple, line 6: what is this used for?

ShawnTurple (approved):

deployments/kustomize/overlays/local/phpmyadmin/deployment.yaml
mhaswell-bcgov: Not sure probably copied from somewhere else. I've removed it.

ShawnTurple (approved):

PR #21, opened 2024-11-04
Feature/descw 2625 naad2 pod

mhaswell-bcgov:


src/Database.php
```diff
@@ -71,11 +71,7 @@ protected function getEntityManager(): EntityManager
     protected function persistAlert(
         Alert $alert
     ): string {
-        try {
-            $this->entityManager->persist($alert);
-        } catch ( UniqueConstraintViolationException $e ) {
-            $this->entityManager = $this->getEntityManager();
-        }
+        $this->entityManager->persist($alert);
```
robrien-bcgov, line 74: how come the try-catch block is removed? is there something about having two deploys that makes it unnecessary?

docker-compose.yaml
```diff
@@ -43,6 +43,7 @@ services:
       MARIADB_SERVICE_PORT: 3306
       NAAD_NAME: NAAD-1
       NAAD_URL: streaming1.naad-adna.pelmorex.com
+      NAAD_REPO_URL: capcp1.naad-adna.pelmorex.com
```
robrien-bcgov, line 46: does docker-compose still work this way? (i'm guessing it just runs one naad-app container at a time)

deployments/kustomize/base/app/job.yaml
```diff
@@ -11,10 +11,6 @@ spec:
           image: bcgovgdx/naad-app
           imagePullPolicy: IfNotPresent
           envFrom:
```
robrien-bcgov, line 13: I'm guessing the database migration doesn't actually need the naad-config and naad-secrets to work. good cleanup

deployments/kustomize/base/app/kustomization.yaml
```diff
@@ -2,17 +2,27 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
     literals:
     - NAAD_NAME=NAAD-1
     - NAAD_URL=streaming1.naad-adna.pelmorex.com
+    - NAAD_REPO_URL=capcp1.naad-adna.pelmorex.com
+  - name: naad2-config
+    literals:
+    - NAAD_NAME=NAAD-2
+    - NAAD_URL=streaming2.naad-adna.pelmorex.com
+    - NAAD_REPO_URL=capcp2.naad-adna.pelmorex.com
+  - name: naad-shared-config
```
robrien-bcgov, line 24: ok this makes sense

robrien-bcgov (approved):
It seems like this would work normally.

The docker-compose shouldn't need to run a 2nd container for the naad-app, so it makes sense to me just to point it at the 1st container

deployments/kustomize/base/app/deployment.yaml
```diff
@@ -1,9 +1,9 @@
 kind: Deployment
 apiVersion: apps/v1
 metadata:
-  name: naad-socket
+  name: naad-socket-1
```
ShawnTurple, line 4: are you sure this is best practice to add the -1 -2, i suspect kubernetes has a better way of doing this, i believe notify does something similar

deployments/kustomize/base/app/kustomization.yaml
```diff
@@ -2,17 +2,27 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - deployment.yaml
+- deployment2.yaml
```
ShawnTurple, line 5: why wouldn't you just add this to the deployment, and use the same config, and just have different env variables

src/Database.php
mhaswell-bcgov: It was moved from being around `persist()` to being around `flush()` because the exception only actually occurs from the call to `flush()`, so it wasn't being caught properly before.

docker-compose.yaml
mhaswell-bcgov: Yeah it still works. Just spawns the one pod instead of 2.

deployments/kustomize/base/app/job.yaml
mhaswell-bcgov: Yeah it only actually needs the database config because it never runs the actual application, just the database migration.

deployments/kustomize/base/app/deployment.yaml
mhaswell-bcgov: Not sure but the pods end up being named pretty clearly which I like:
```
NAME                             READY   STATUS      RESTARTS       AGE
naad-socket-1-644cb589f7-jx6vj   1/1     Running     70 (11m ago)   3d1h
naad-socket-2-69955dc986-tq58d   1/1     Running     64 (78m ago)   3d1h
```

deployments/kustomize/base/app/kustomization.yaml
mhaswell-bcgov: I'm not sure how to make 2 pods spawn from the one deployment and have them use different configMaps to set the env variables. Open to suggestions so we're doing less duplication.

ShawnTurple: you have 2 deployments in the deployments.yaml, and both use the same config map, however one uses one set of env, and the other uses the 2nd set

mhaswell-bcgov: I think that's what's happening now except they're split into 2 files. The env variables that are different between 1 and 2 come from two different config maps and the env variables that are the shared come from one shared config map.

PR #22, opened 2024-11-14
DESCW-2704 remove rss tests and tweak test output

robrien-bcgov:
- remove NaadRSSClient class
- remove get_rss module that uses it
- Remove rss tests 
- update phpunit to v11
- and tweak test output format with `--testdox` option
recommended in https://phpunit.de/getting-started/phpunit-11.html
image


composer.json
```diff
@@ -33,7 +33,7 @@
             "vendor/bin/phpcbf ./src"
         ],
         "test": [
-            "vendor/bin/phpunit ./tests"
+            "vendor/bin/phpunit --testdox ./tests"
```
mhaswell-bcgov, line 36: I like this. More descriptive than just the dots it usually shows.

mhaswell-bcgov (changes requested):
There are a few more files that can be deleted: `src/get_rss.php` and `src/NaadRssClient.php`.

composer.json
robrien-bcgov: i like it too.

this pr is just for the test though - I was planning to remove the other rss stuff in the next pr.

lately i'm into smal prs with one theme

robrien-bcgov: i like it too.

this pr is just for the test though - I was planning to remove the other rss stuff in the next pr.

lately i'm into smal prs with one theme

mhaswell-bcgov (approved):

composer.json
mhaswell-bcgov: Small PRs are ok but we don't have a way to indicate that the PR doesn't cover the whole ticket. Not sure how we can do that but it would be helpful for reviewers.

robrien-bcgov: yeah good point. I guess the best I can do is make a comment in the pr description saying "this is just for X, the rest of the changes are coming in another PR" or something like that

PR #23, opened 2024-11-15
Feature/descw 2739 refactor passing environment variables to php

robrien-bcgov:
This small pr will refactor Dockerfile to avoid using a password.

It leverages a secret mount, and extracts that password from your local .env file

This pr also includes updated Docker build/run instructions in the README

Lastly, The scripts which build the image and run it (composer.json) have been altered to work properly. They provide a simple way to build an image and run it:

```bash
#build only
composer build

#build and start the app (docker compose handles the 'swarm' of running containers)
composer start
```

PS. this is just the first requirement of the ticket, and that is why i'm doing a PR onto a feature branch, with more small PRs to follow.


Dockerfile
```diff
@@ -1,29 +1,51 @@
...
-COPY ./ /var/www/html/
-COPY ./entrypoint.sh /home/
-RUN touch $LOG_FILE_PATH && chown 1001 $LOG_FILE_PATH
+# Set environment variables
+ENV NAAD_NAME=NAAD-1 \
+NAAD_URL=streaming1.naad-adna.pelmorex.com \
+DESTINATION_URL=https://localhost/wp-json/naad/v1/alert \
+DESTINATION_USER=naadbot \
+LOG_FILE_PATH="/var/www/html/naad-socket.log" \
+DATABASE_NAME="naad_connector"
```
mhaswell-bcgov, line 20: Sorry I thought I had already deleted these in a previous commit but I didn't. It should be possible to delete all of these except `LOG_FILE_PATH` and allow these values come from `docker-compose.yaml`/`.env` or the `kustomization.yaml` files depending on environment.

composer.json
```diff
@@ -20,11 +20,14 @@
     },
     "scripts": {
         "build": [
-            "docker build -t naad-app ."
+            "docker build --secret id=destination_password,src=.env -t bcgovgdx/naad-app ."
```
mhaswell-bcgov, line 23: Don't think we need this if we remove it from the Dockerfile.

mhaswell-bcgov (changes requested):
I think we can remove most of the env variables from the Dockerfile and then we won't have to use the `--secret` flag in the build command.

composer.json
robrien-bcgov: it has to be in the dockerfile as a secret mount (it extracts the password from your .env file inside the container)

```dockerfile
# Use Docker secret extracted from your .env file.
RUN --mount=type=secret,id=destination_password \
  destination_password=$(cat /run/secrets/destination_password) && \
  echo "destination_password=$destination_password" > /var/www/html/.env
 ```

Dockerfile
robrien-bcgov: Good idea. I'll play around with this on Monday

composer.json
robrien-bcgov: my `docker compose` only uses the image, but doesn't build it, so this is necessary for the docker compose deploy, I think

The kubernetes deploy probably doesn't need it though - I'll test that on Monday

PR #24, opened 2024-11-15
Refactored NaadSocketClient:

mhaswell-bcgov:
- Moved socket connection logic to new NaadSocketConnection class
    - Moved fetching of missed alerts from NAAD alert repository to NaadRepositoryClient class Refactored CustomLogger to extend Monolog/Logger to allow us to create test double Refactored NaadSocketClientTest to test only the public function handleResponse() instead of using reflection class to test internal functions Updated phpunit.xml to correctly configure phpunit to run code coverage reports Added test-coverage command to composer.json
Fixed issues with app deployments missing the database name env variable (was renamed from DATABASE_NAME to MARIADB_DATABASE so it could be reused between all pods)

Note: we'll still need to update the NaadSocketClient unit tests as they don't have full coverage and are not testing interactions with the test doubles.


src/NaadSocketClient.php
```diff
@@ -43,143 +35,40 @@ class NaadSocketClient
...
+     *                                                NaadRepositoryClient.
      */
     public function __construct(
         string $name,
-        string $socketUrl,
         DestinationClient $destinationClient,
         Logger $logger,
         Database $database,
-        int $port = 8080,
+        NaadRepositoryClient $repositoryClient = null
```
ShawnTurple, line 65: i'm guessing you can't use the default as a new NaadRepositoryClient(), as this would save you a little on line 71

src/NaadSocketConnection.php
```diff
@@ -0,0 +1,126 @@
...
+        $this->logger->info('Creating socket');
+        $socket = socket_create(AF_INET, SOCK_STREAM, SOL_TCP);
+        if ($socket === false ) {
+            $this->logger->error(
+                'socket_create() failed: reason: {error}',
+                [
+                    'error' => socket_strerror(socket_last_error()),
+                ]
+            );
+            return 2;
```
ShawnTurple, line 74: if we are returning an integer, and it is not obvious, it might be helpful to indicate why we are returning 2

src/NaadSocketConnection.php
```diff
@@ -0,0 +1,126 @@
...
+        } else {
+            $this->logger->info('OK.');
```
ShawnTurple, line 76: not a really helpful message... maybe 
```suggestion
            $this->logger->info('Socket OK.');
```

src/NaadSocketConnection.php
```diff
@@ -0,0 +1,126 @@
...
+        if ($result === false ) {
+            $error = socket_strerror(socket_last_error($socket));
+            $this->logger->error(
+                "socket_connect() failed.\nReason: ({result}) {error}",
+                [
+                    'result' => $result,
+                    'error'  => $error,
+                ]
+            );
+            return 3;
```
ShawnTurple, line 99: again knowing why we are returning 3 might be useful

ShawnTurple (approved):

src/NaadSocketClient.php
mhaswell-bcgov: Yeah you can't default to that probably because it's possible for a constructor to throw an exception which you wouldn't be able to catch in a default argument assignment.

src/NaadSocketConnection.php
mhaswell-bcgov: It's an exit code but doesn't really have a meaning that would be used anywhere. I mostly just wanted to make sure it returned a non-zero value so that it would indicate to Kubernetes that the pod exited due to failure rather than it having been stopped on purpose. Not sure what best practice is here.

src/NaadSocketConnection.php
mhaswell-bcgov: I've updated the log messages in this file to be more descriptive.

src/NaadSocketConnection.php
ShawnTurple: then a comment saying this returning a non zero, and if they are just 0 or non zero we might want to keep consistent in the number

ShawnTurple (approved):

PR #28, opened 2024-11-19
DESCW-2669 Add error handling for missing identifiers in alerts

kshapka-bcgov:
- Adds error handling for alerts with missing or empty identifier fields
  - https://citz-gdx.atlassian.net/browse/DESCW-2669
- Adds colours flag to testing for better visibility.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
             "vendor/bin/phpcbf ./src"
         ],
         "test": [
-            "vendor/bin/phpunit --testdox"
+            "vendor/bin/phpunit --testdox --colors=auto"
         ],
         "test-coverage": [
             "vendor/bin/phpunit --testdox --coverage-html=coverage/"
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
             "vendor/bin/phpcbf ./src"
         ],
         "test": [
-            "vendor/bin/phpunit --testdox --colors=auto"
+            "vendor/bin/phpunit"
         ],
         "test-coverage": [
             "vendor/bin/phpunit --testdox --coverage-html=coverage/"
--- a/src/Entity/Alert.php
+++ b/src/Entity/Alert.php
@@ -5,6 +5,7 @@ namespace Bcgov\NaadConnector\Entity;
 use DateTime;
 use Doctrine\ORM\Mapping as ORM;
 use SimpleXMLElement;
+use Exception;
 
 #[ORM\Entity]
 #[ORM\Table(name: 'alerts')]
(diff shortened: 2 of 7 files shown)


src/Entity/Alert.php
```diff
@@ -184,16 +185,29 @@ public function setSuccess( bool $success ): void
...
+        try {
+            $alert = new Alert();
+            $alert->setId($identifier);
+            $alert->setBody($xml->asXML());
+            $alert->setReceived(new DateTime());
+    
+            return $alert;
+        } catch (Exception $e) {
+            // Log meaningful error message and rethrow the exception.
+            error_log('Failed to create Alert from XML: ' . $e->getMessage());
```
mhaswell-bcgov, line 209: I wonder if we need to error_log() in the Alert class or whether we should change the try/catch in the socket client class and let its Logger do it.

src/NaadSocketClient.php
```diff
@@ -132,14 +132,17 @@ public function handleResponse( string $response ): bool
      */
     protected function insertAlert( SimpleXMLElement $xml )
     {
+        $alert = null;
+
         try {
             $alert = Alert::fromXml($xml);
```
mhaswell-bcgov, line 138: Re: my other comment, here we'd have:
```
try {
    $alert = Alert::fromXml($xml);
} catch(Exception $e) {
    // Log the error specific to failing to get an Alert from the XML using the Logger.
}
try {
    $this->database->insertAlert($alert);
} catch ...
    // Log an error specific to failing to insert into the database using the Logger.
```
Because right now it isn't super clear from the log where the exception actually occurred.

composer.json
```diff
@@ -33,7 +33,7 @@
             "vendor/bin/phpcbf ./src"
         ],
         "test": [
-            "vendor/bin/phpunit --testdox"
+            "vendor/bin/phpunit --testdox --colors=auto"
```
mhaswell-bcgov, line 36: I like this. Is it possible to add it to the `phpunit.xml` somehow so this (and --testdox while we're at it) gets applied by default when just running the `phpunit` command?

mhaswell-bcgov (approved):

composer.json
mhaswell-bcgov: I tried it because I was curious. This seems to work in phpunit.xml:
```
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         ...
         colors="true"
         testdox="true">
```

src/Entity/Alert.php
```diff
@@ -205,8 +203,6 @@ public static function fromXml( SimpleXMLElement $xml ): Alert
     
             return $alert;
         } catch (Exception $e) {
-            // Log meaningful error message and rethrow the exception.
```
mhaswell-bcgov: Thanks. I don't think we need the try/catch here anymore since we're just throwing the error again anyway.

mhaswell-bcgov (approved):

PR #29, opened 2024-11-20
DESCW-2764 Replace raw curl with guzzle in DestinationClient

kshapka-bcgov:
- Replaces raw curl with guzzle in DestinationClient.
- Removes try/catch in Alert which was already being handled elsewhere.


robrien-bcgov (approved):
nice. much simpler

PR #30, opened 2024-11-20
Added react/socket dependency to composer.json

mhaswell-bcgov:
Created src/start_socket_server.php script to start a socket using react/socket library


robrien-bcgov (approved):
nice and simple

PR #33, opened 2024-11-26
descw-2739 update Kustomize and Readme

robrien-bcgov:
- ([DESCW-2739](https://citz-gdx.atlassian.net/browse/DESCW-2739))
- test and update K8s deployment so it works properly with NaadVars class getter.
- fix typo in socket client NaadVars.
- Update Readme layout and instructions.
- add scripts to simplify k8s build/start/stop for local development.
- add socket client try/catch block to catch environment variable misconfiguration.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,8 +27,16 @@
             "Composer\\Config::disableProcessTimeout",
             "docker compose up"
         ],
+        "startPods": [
+            "composer build",
+            "kubectl config use-context docker-desktop",
+            "kubectl apply -k deployments/kustomize/overlays/local"
+        ],
         "stop": [
             "docker compose down"
+        ],
+        "stopPods": [
+            "kubectl delete -k deployments/kustomize/overlays/local"
         ],
         "phpcs": [
             "# show the rule that is being applied using the -s flag",
--- a/deployments/kustomize/base/app/kustomization.yaml
+++ b/deployments/kustomize/base/app/kustomization.yaml
@@ -2,8 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - deployment.yaml
-- deployment2.yaml
 - job.yaml
+- deployment2.yaml
 
 # Updates the labels for all containers.
 commonLabels:
--- a/src/NaadSocketClient.php
+++ b/src/NaadSocketClient.php
@@ -92,7 +92,15 @@ class NaadSocketClient
 
         if ($this->isHeartbeat($xml) ) {
             $this->logger->info('Heartbeat received.');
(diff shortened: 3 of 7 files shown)


composer.json
```diff
@@ -27,9 +27,17 @@
             "Composer\\Config::disableProcessTimeout",
             "docker compose up"
         ],
+        "startPods": [
```
mhaswell-bcgov, line 30:
```suggestion
        "k8s-start": [
```
Might be a clearer name because the deployment creates more than just pods in Kubernetes.

composer.json
```diff
@@ -27,9 +27,17 @@
...
+            "composer build",
+            "kubectl config use-context docker-desktop",
+            "kubectl apply -k deployments/kustomize/overlays/local"
+        ],
         "stop": [
             "docker compose down"
         ],
+        "stopPods": [
```
mhaswell-bcgov, line 38:
```suggestion
        "k8s-delete": [
```

deployments/kustomize/base/app/kustomization.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
 - deployment.yaml
-- deployment2.yaml
 - job.yaml
+- deployment2.yaml
```
mhaswell-bcgov, line 6: I don't think this actually changes the order that they start up. From what I understand there's no way to control the  order resources start up in Kubernetes. You're supposed to just make everything as resilient to failures as possible so they can recover on their own and sort themselves out.

src/NaadSocketClient.php
```diff
@@ -92,7 +92,15 @@ public function handleResponse( string $response ): bool
 
         if ($this->isHeartbeat($xml) ) {
             $this->logger->info('Heartbeat received.');
-            $missedAlerts = $this->findMissedAlerts($xml);
+            try {
+                $missedAlerts = $this->findMissedAlerts($xml);
```
mhaswell-bcgov, line 96: I think the only Exception that can be thrown here is from the call to `$this->database->getAlertsById()` inside `findMissedAlerts()` so this try catch would be better placed around that call.

mhaswell-bcgov (changes requested):

deployments/kustomize/base/app/kustomization.yaml
robrien-bcgov: but it works though

src/NaadSocketClient.php
robrien-bcgov: kk i'll refactor this tomorrow

mhaswell-bcgov (approved):

PR #37, opened 2024-12-02
DESCW-2771 replace curl with guzzle client

robrien-bcgov:
4 commits squashed using `git rebase -i HEAD~4`

- still needs cleanups and finishing touches. DESCW-2771 adds guzzle instead of CURL for grabbing the alerts from the naad repo
- some minor refactoring as well, for easy maintenance descw-2771 fix typo
descw-2771 slightly refactor the socket client to fix failing tests

### DIFF
--- a/src/NaadRepositoryClient.php
+++ b/src/NaadRepositoryClient.php
@@ -1,7 +1,8 @@
 <?php
 namespace Bcgov\NaadConnector;
 use Bcgov\NaadConnector\NaadVars;
-
+use GuzzleHttp\Client;
+use GuzzleHttp\Exception\RequestException;
 /**
  * NaadRepositoryClient class makes requests to the NAAD alert repository.
  *
@@ -13,6 +14,13 @@ use Bcgov\NaadConnector\NaadVars;
  */
 class NaadRepositoryClient
 {
+
+    /**
+     * Guzzle HTTP client.
+     *
+     * @var Client
+     */
+    protected Client $client;
 
     /**
      * The url of the NAAD alert repository.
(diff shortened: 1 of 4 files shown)


src/NaadSocketClient.php
```diff
@@ -168,47 +168,80 @@ protected function insertAlert( SimpleXMLElement $xml )
      * @return bool|SimpleXMLElement Returns SimpleXMLElement on success,
      *                               false otherwise.
      */
-    protected function validateResponse( string $response ): bool|SimpleXMLElement
+    protected function validateResponse(?string $response): bool|SimpleXMLElement
```
mhaswell-bcgov, line 171: Why is this change needed?

robrien-bcgov: the tests fail on this method because they send a null instead of a string.

that's why i changed the type hinting to allow a null to be sent

robrien-bcgov: look at this - it's failing because the method expects a string, not null

It only happens on this branch
image

src/NaadRepositoryClient.php
```diff
@@ -31,54 +39,65 @@ public function __construct()
...
-     * @param string $s The string to perform the replacement on.
+     * @param array $reference Heartbeat references array parts (sender, id, sent).
      *
-     * @return string
+     * @return string|null The alert response body or null if an error occurs.
+     *
+     * @throws Exception if an error occurs during the GET request.
      */
-    protected function replaceUrlCharacters( string $s ): string
+    public function fetchAlert(array $reference): ?string
```
mhaswell-bcgov, line 50:
```suggestion
    public function fetchAlert(string $reference): ?string
```
It's still a string at this point isn't it?

src/NaadRepositoryClient.php
```diff
@@ -31,54 +39,65 @@ public function __construct()
...
+        $url = $this->getURL($reference);
+        $this->client = new Client();
+
+        try {
+            $response = $this->client->get($url);
+
+            return (string) $response->getBody();
+        } catch (RequestException $e) {
+            error_log("Request failed: " . $e->getMessage());
+            return null;
```
mhaswell-bcgov, line 61: Can we have this throw the Exception instead of returning null so we don't have to change the SocketClient::validateAlert() arguments?

mhaswell-bcgov (changes requested):

src/NaadRepositoryClient.php
```diff
@@ -31,54 +39,65 @@ public function __construct()
...
+     * @return string The constructed URL.
      */
-    public function fetchAlert(
-        array $reference,
-    ): string {
-        $url = sprintf(
+    public function getURL(array $reference)
+    {
+        // Extract and sanitize values from the reference array.
+        $date = strtok($reference['sent'], 'T');
```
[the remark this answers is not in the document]
robrien-bcgov: yeah the url will be me3ssed up, teh request will fail in the try/catch block, and fetchAlert will return null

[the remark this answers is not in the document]
robrien-bcgov: or in this case, it will throw an error

mhaswell-bcgov (approved):

mhaswell-bcgov (approved):

PR #39, opened 2024-12-04
descw-2792 add .dockerignore

robrien-bcgov:


Dockerfile
```diff
@@ -1,36 +1,33 @@
...
 
-# Install linux package dependencies and PHP extensions
-RUN set -ex; \
-  apt update && apt install -y libzip-dev && \
-  docker-php-ext-configure zip && \
-  docker-php-ext-install sockets zip pdo pdo_mysql && \
-  docker-php-ext-enable sockets && \
-  apt-get clean && \
-  rm -rf /var/lib/apt/lists/*
+# Install dependencies and PHP extensions
```
ShawnTurple, line 3: in future, only the acceptance criteria, if you think the dockerfile needs to be cleaned up, then it should be discussed, and a ticket created, otherwise it makes a simple pr into one that we have to understand what you actually did, and if it has any impact,

.dockerignore
```diff
@@ -0,0 +1,28 @@
...
+openshift
+phpcs.xml
+phpunit.xml
+sample-env
+tests
+vendor
+.git
+node_modules
+Dockerfile
+docker-compose.yaml
```
mhaswell-bcgov, line 17:
```suggestion
docker-compose.yaml
.devcontainer
.DS_Store
.env
.github
.gitignore
.phpunit.cache
.phpunit.result.cache
```
I think we could exclude even a few more although you might not have some of these files so they wouldn't show up for you.
image

mhaswell-bcgov (changes requested):

mhaswell-bcgov (approved):

PR #44, opened 2024-12-09
descw-2799 adds unit tests for the naadvars class.

robrien-bcgov:
- uses a local .env.test in the /test folder to mock environment vars
- will require composer install to catch up to the changes (adds dotenv package)

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
       "vendor/bin/phpunit"
     ],
     "test-coverage": [
-      "vendor/bin/phpunit --coverage-html=coverage/"
+      "vendor/bin/phpunit --coverage-text"
     ],
     "migrate": [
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -52,7 +52,7 @@
       "vendor/bin/phpunit"
     ],
     "test-coverage": [
-      "vendor/bin/phpunit --coverage-html=coverage/"
+      "vendor/bin/phpunit --coverage-text"
```
robrien-bcgov, line 55: @mhaswell-bcgov Just adding this option because it works and is useful for testing if my class has coverage or not.

I'm not sure what the former script was meant to do, but I can't see any new .html files when I ran it.

Maybe we should just have another script that writes an html report or something fancier.

kshapka-bcgov (approved):

PR #47, opened 2024-12-12
descw-2795 adds tests for the Custom Logger

robrien-bcgov:
### DEC 12, 2024

- Add tests for CustomLogger class [DESCW-2795](https://citz-gdx.atlasssian.net/browse/DESCW-2795)
  - implemented suggested changes
image

### DIFF
--- a/tests/CustomLoggerTest.php
+++ b/tests/CustomLoggerTest.php
@@ -1,0 +1,231 @@
+<?php declare(strict_types=1);
+
+use PHPUnit\Framework\Attributes\{
+  Test,
+  CoversClass,
+  DataProvider,
+  UsesClass,
+};
+
+use PHPUnit\Framework\TestCase;
+use Monolog\Level;
+use Monolog\Handler\StreamHandler;
+use Bcgov\NaadConnector\CustomLogger;
+
+/**
+ * CustomLogger Class for testing CustomLogger class.
+ * This will test the class constructor, and _convertLogLevel
+ * which is a private method
+ *
+ * @category Client
+ * @package  NaadConnector
+ * @author   Richard O'Brien <Richard.OBrien@gov.bc.ca>
+ * @license  https://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link     https://alerts.pelmorex.com/
+ */
+#[CoversClass('Bcgov\NaadConnector\CustomLogger')]
+#[UsesClass('Bcgov\NaadConnector\CustomLogger')]
+final class CustomLoggerTest extends TestCase
+{
+
+    /**
+     * Test the CustomLogger Constructor.
+     *
+     * @return void
+     */
+    #[Test]
+    public function testDefaultConstructor()
+    {
+        $logger = new CustomLogger();
+
+        // Check if the handlers are configured as expected.
+        $handlers = $logger->getHandlers();
+        $this->assertCount(1, $handlers);
+        $this->assertInstanceOf(StreamHandler::class, $handlers[0]);
+
+        // Assert the default logging level is 'info'
+        $streamHandler = $handlers[0];
+        $this->assertEquals(Level::Info, $streamHandler->getLevel());
+
+        // Assert that the default channel name is 'monolog'.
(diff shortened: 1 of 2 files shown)


tests/CustomLoggerTest.php
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+
+    /**
+     * Returns a mapping of log level strings to their corresponding
+     * Monolog Level constants.
+     *
+     * @return array<string, int>
+     */
+    private function getLogLevelMappings()
```
mhaswell-bcgov, line 105: This could be done using a PHPUnit DataProvider (see https://github.com/bcgov/naad-connector/blob/main/tests/NaadSocketClientTest.php#L51 for an example).

tests/CustomLoggerTest.php
```diff
@@ -0,0 +1,231 @@
...
+     *
+     * @return void
+     */
+    #[Test]
+    public function testLogLevelConversion()
+    {
+        $logLevels = $this->getLogLevelMappings();
+
+        foreach ($logLevels as $levelString => $expectedLevel) {
+            $logger = $this->createLoggerWithLevel($levelString);
```
mhaswell-bcgov, line 88:
```suggestion
            $logger = new CustomLogger('test', $levelString);
```
Might be simpler, especially since now we have to write PHPDoc comments for every function here.

tests/CustomLoggerTest.php
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * Test invalid log level.
+     * Expect that an invalid logging level string will cause
+     * the level to be set to 'info'.
+     *
+     * @return void
+     */
+    #[Test]
+    public function testInvalidLogLevel()
```
mhaswell-bcgov, line 155: This could be a case in the DataProvider above:
```
    'invalidLevel' => Level::info
```

tests/CustomLoggerTest.php
```diff
@@ -0,0 +1,231 @@
...
+        $this->assertEquals(Level::Info, $streamHandler->getLevel());
+    }
+
+    /**
+     * Test logger output matches expected output
+     *
+     * @return void
+     */
+    #[Test]
+    public function testLoggerOutput()
```
mhaswell-bcgov, line 171: I don't think we should test this as this is more of a test of the Monolog\Logger class than our own code.

mhaswell-bcgov (changes requested):

tests/CustomLoggerTest.php
robrien-bcgov: done
image

tests/CustomLoggerTest.php
robrien-bcgov: Good idea, it's now a test case in the data provider

tests/CustomLoggerTest.php
robrien-bcgov: removed - it did seem like a messy test

tests/CustomLoggerTest.php
mhaswell-bcgov: Beautiful

tests/CustomLoggerTest.php
```diff
@@ -0,0 +1,129 @@
...
+        string $levelString,
+        Monolog\Level $expectedLevel
+    ): void {
+        $logger = new CustomLogger('test', $levelString);
+        $actualLevel = $logger->getHandlers()[0]->getLevel();
+
+        $this->assertEquals(
+            $expectedLevel,
+            $actualLevel,
+            <<<MESSAGE
```
mhaswell-bcgov, line 119: What is the message if you don't include this? I assume it's pretty close to this by default.

mhaswell-bcgov (approved):

tests/CustomLoggerTest.php
robrien-bcgov: It's similar, and slightly redundant.  
Maybe it's better not to include it, though I like the clarity of a multiline message. I can remove it in my next pr if you like

image

PR #48, opened 2024-12-12
DESCW-2318 DestinationClientTest improvements

kshapka-bcgov:
It was necessary to refactor DestinationClient slightly in order to maximize code coverage:
DestinationClient is now passed the logger and (optionally) a client.
This allows the elimination of reflection classes and 100% code coverage which was otherwise blocked by outer try/catch scenarios.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,9 @@
     "test-coverage": [
       "vendor/bin/phpunit --coverage-text"
     ],
+    "test-report": [
+      "vendor/bin/phpunit --coverage-html=coverage/"
+    ],
     "migrate": [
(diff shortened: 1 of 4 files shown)


robrien-bcgov (approved):
LGTM

Interesting pattern, passing in the logger like that.
I guess it makes testing better somehow

kshapka-bcgov:
> LGTM
> 
> Interesting pattern, passing in the logger like that. I guess it makes testing better somehow

Yes - covering some exception cases were impossible as the code was written, because some lines were always bypassed by outer try/catch blocks.

PR #49, opened 2024-12-17
Add tests for NaadRepositoryClient

robrien-bcgov:
- ([DESCW-2797](https://citz-gdx.atlassian.net/browse/DESCW-2797))

### DIFF
--- a/src/NaadRepositoryClient.php
+++ b/src/NaadRepositoryClient.php
@@ -78,16 +78,12 @@ class NaadRepositoryClient
             );
         }
 
-        $date          = strtok($reference['sent'], 'T');
-        $sanitizedSent = strtr($reference['sent'], self::SANITIZE_RULES);
-        $sanitizedId   = strtr($reference['id'], self::SANITIZE_RULES);
-
         return sprintf(
             self::URL_TEMPLATE,
             $this->baseUrl,
-            $date,
-            $sanitizedSent,
-            $sanitizedId
+            strtok($reference['sent'], 'T'), // date
+            strtr($reference['sent'], self::SANITIZE_RULES), // sent
+            strtr($reference['id'], self::SANITIZE_RULES), // id
         );
     }
 
--- a/src/NaadRepositoryClient.php
+++ b/src/NaadRepositoryClient.php
@@ -70,14 +70,8 @@ class NaadRepositoryClient
      *
      * @return string The constructed URL.
      */
-    public function constructURL(array $reference): string
+    protected function constructURL(array $reference): string
     {
-        if (!isset($reference['sent'], $reference['id'])) {
-            throw new \InvalidArgumentException(
-                "Reference must contain 'sent' and 'id' keys"
-            );
-        }
-
         return sprintf(
             self::URL_TEMPLATE,
             $this->baseUrl,
--- a/src/NaadRepositoryClient.php
+++ b/src/NaadRepositoryClient.php
@@ -14,6 +14,13 @@ use GuzzleHttp\Exception\RequestException;
  */
 class NaadRepositoryClient
 {
+    /**
+     * Constants used by these methods
+     * - Url template for generating an URL string based on inputs
+     * - Sanitize rules to translate a reference to an URL-friendly string
+     */
+    protected const URL_TEMPLATE = 'http://%s/%s/%sI%s.xml';
+    protected const SANITIZE_RULES = ['-' => '_', '+' => 'p', ':' => '_'];
 
     /**
      * Guzzle HTTP client.
@@ -30,11 +37,19 @@ class NaadRepositoryClient
     protected string $baseUrl;
 
     /**
-     * Constructor for NaadRepositoryClient.
+     * Initializes a new instance of the NaadRepositoryClient class.
+     *
+     * @param Client|null   $client   The Guzzle HTTP client to use.
+     *                                Defaults to a new Client instance.
+     * @param NaadVars|null $naadVars The NaadVars instance to use.
+     *                                Defaults to a new NaadVars instance.
      */
-    public function __construct()
-    {
-        $naadVars       = new NaadVars();
+    public function __construct(
+        Client $client     = null,
+        NaadVars $naadVars = null
+    ) {
+        $this->client   = $client ?: new Client();
+        $naadVars       = $naadVars ?: new NaadVars();
         $this->baseUrl  = $naadVars->naadRepoUrl;
     }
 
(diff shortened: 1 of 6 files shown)


src/NaadRepositoryClient.php
```diff
@@ -69,33 +83,46 @@ public function fetchAlert(array $reference): string
...
      *
-     * @return string The modified string.
+     * @param string $property The name of the property to retrieve.
+     *
+     * @return string The value of the requested property.
+     *
+     * @throws \InvalidArgumentException If the property does not exist.
      */
-    protected function replaceUrlCharacters(string $s): string
+    public function __get($property): string
```
mhaswell-bcgov, line 118: Where is this used?

src/NaadRepositoryClient.php
```diff
@@ -69,33 +83,46 @@ public function fetchAlert(array $reference): string
      *
      * @return string The constructed URL.
      */
-    protected function getURL(array $reference)
+    public function constructURL(array $reference): string
```
mhaswell-bcgov, line 86: I like the name change but I don't think this should be public unless there's a need for this url to be used somewhere else in the application.

src/NaadRepositoryClient.php
```diff
@@ -69,33 +83,46 @@ public function fetchAlert(array $reference): string
...
     {
-        // Extract and sanitize values from the reference array.
+        // Extract date and sanitize values for URL construction.
         $date = strtok($reference['sent'], 'T');
-        $sent = $this->replaceUrlCharacters($reference['sent']);
-        $id = $this->replaceUrlCharacters($reference['id']);
+        $sanitize = fn(string $value): string => strtr(
```
mhaswell-bcgov, line 90: To me this is less readable than giving the sprintf() `strtr($reference['sent'], self::SANITIZE_RULES)` for example.

src/NaadRepositoryClient.php
```diff
@@ -30,11 +37,19 @@ class NaadRepositoryClient
...
+     * @param Client|null   $client   The Guzzle HTTP client to use.
+     *                                Defaults to a new Client instance.
+     * @param NaadVars|null $naadVars The NaadVars instance to use.
+     *                                Defaults to a new NaadVars instance.
      */
-    public function __construct()
-    {
-        $naadVars       = new NaadVars();
+    public function __construct(
+        Client $client     = null,
```
mhaswell-bcgov, line 48: I don't think this should be optional as it feels like the way the tests are using this class and the way it's used in the actual application don't match up. Making it required means we'll also have to change src/start.php to pass an instance of NaadRepositoryClient into the NaadSocketClient.

src/NaadRepositoryClient.php
```diff
@@ -30,11 +37,19 @@ class NaadRepositoryClient
...
+        NaadVars $naadVars = null
+    ) {
+        $this->client   = $client ?: new Client();
+        $naadVars       = $naadVars ?: new NaadVars();
```
mhaswell-bcgov, line 52: If NaadVars is only used to set baseUrl then we could probably make it so this class takes the baseUrl as an argument.

mhaswell-bcgov (dismissed):

src/NaadRepositoryClient.php
robrien-bcgov: This is dependency injection hack and I need it for the test.

What did you want instead, exactly?

mhaswell-bcgov: We're never actually constructing a NaadRepositoryClient because we're doing the same optional argument thing here: https://github.com/bcgov/naad-connector/blob/main/src/NaadSocketClient.php#L77 so what I'd want to see is making that argument also required. Then we'd create an instance of NaadRepositoryClient in start.php like we do with the other classes and pass it into the NaadSocketClient instance.

src/NaadRepositoryClient.php
robrien-bcgov: Not a bad idea!

I guess if we're only using this class as a dependency for NaadSocketClient, and since NaadSocketClient only uses teh baseUrl from an instance of NaadVars, It makes sense to simplify this.

Of course my tests would have to be refactored, but I was finding the tests a little sus anyway

src/NaadRepositoryClient.php
robrien-bcgov: I think I needed this public for testing the method in isolation, but I think I could set it protected. I'll try it.

src/NaadRepositoryClient.php
robrien-bcgov: fair enough - was looking to save code lines, but readability is more important. I'll fix it.

src/NaadRepositoryClient.php
robrien-bcgov: This is used in a test (it's a magic getter)

```php
    #[Test]
    public function testConstructorInitializesBaseUrl(): void
    {
        $this->naadVarsMock->naadRepoUrl = 'https://mock-naad-repo.com';

        $client = new NaadRepositoryClient(null, $this->naadVarsMock);

        $this->assertSame('https://mock-naad-repo.com', $client->baseUrl);
    }
    ```

File: src/NaadRepositoryClient.php (hunk 4 above)
Comment (robrien-bcgov) [reply src/NaadRepositoryClient.php:48]: we need to refactor the socket client to inject the dependent class instance.

2 places
in socket client
in start script

File: tests/NaadRepositoryClientTest.php [hunk 6]
```diff
@@ -0,0 +1,216 @@
...
+    #[Test]
+    #[DataProvider('referenceData')]
+    public function testFetchAlert(array $reference): void
+    {
+        $this->naadVarsMock->naadRepoUrl = 'mock-repo-url.com';
+        $this->guzzleClientMock
+            ->method('get')
+            ->willReturn(new Response(200, [], 'Mock Alert Response'));
+
+        $client = new NaadRepositoryClient(
```
robrien-bcgov, line 103: test constructURL function implicitly.

expect the url to be <expected url>

src/NaadRepositoryClient.php
robrien-bcgov: check if the method that uses baseUrl was claled with the expected base url

robrien-bcgov: https://github.com/bcgov/naad-connector/blob/main/tests/DestinationClientTest.php#L121

src/NaadRepositoryClient.php
```diff
@@ -50,11 +65,10 @@ public function __construct()
     public function fetchAlert(array $reference): string
     {
         try {
```
robrien-bcgov, line 67: move the try to only encompass the get($url);

check if the other lines (currently in the block) are able to throw an exception.

mhaswell-bcgov (approved):
Looks good! Clean and simple.

PR #51, opened 2025-01-10
Added Kustomize overlay for OpenShift deployments

mhaswell-bcgov:
DESCW-2848, DESCW-2818

Added network policy to allow all traffic between pods in the namespace.
Fixed bug when updating alerts in DestinationClient causing out of range error on failures column.
Fixed file permissions bug in OpenShift on heartbeat.log file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ RUN chmod +x /app/entrypoint.sh
 RUN useradd -ms /bin/bash appuser \
     && mkdir /app/vendor \
     && touch /app/heartbeat.log \
+    && chmod 777 /app/heartbeat.log \
     && chown -R appuser:appuser /app
 
 # Switch to non-root user
--- a/deployments/kustomize/overlays/openshift/network-policy.yaml
+++ b/deployments/kustomize/overlays/openshift/network-policy.yaml
@@ -1,0 +1,12 @@
+# TODO: Replace with specific network policies instead of allowing all traffic.
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: allow-same-namespace
+spec:
+  podSelector: {}
+  ingress:
+    - from:
+        - podSelector: {}
+  policyTypes:
+    - Ingress
--- a/src/Database.php
+++ b/src/Database.php
@@ -122,10 +122,12 @@ class Database
         $query = $qb->select(array('a'))
             ->from(Alert::class, 'a')
             ->where('a.success = false')
-            ->andWhere($qb->expr()->orX(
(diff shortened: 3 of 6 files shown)


deployments/kustomize/overlays/openshift/network-policy.yaml
```diff
@@ -0,0 +1,12 @@
+# TODO: Replace with specific network policies instead of allowing all traffic.
```
ShawnTurple, line 1: please create the ticket for this, and next change update the todo with the ticket ref

Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN chmod +x /app/entrypoint.sh
 RUN useradd -ms /bin/bash appuser \
     && mkdir /app/vendor \
     && touch /app/heartbeat.log \
+    && chmod 777 /app/heartbeat.log \
```
ShawnTurple, line 29: this also would be a temporary solution, i though we were going the route of [emptydir](https://citz-gdx.atlassian.net/browse/SDE-20528)

deployments/kustomize/overlays/openshift/network-policy.yaml
mhaswell-bcgov: Ticket that should fix this is already in the current sprint: https://citz-gdx.atlassian.net/browse/DESCW-2752

src/Database.php
```diff
@@ -88,34 +72,20 @@ protected function persistAlert(
     public function insertAlert( Alert $alert ): void
     {
         try {
-            $this->persistAlert($alert);
-            $this->entityManager->flush();
+            $this->entityManager->persist($alert);
```
ShawnTurple, line 75: its not obvious to me how this is related to OpenShift deployments, what problem are we solving here?

src/Database.php
```diff
@@ -144,25 +114,27 @@ public function getUnsentAlerts(): array
...
-            )'
-        );
-
         // Calculate the retry threshold based on failure count.
         $retryThreshold = clone $now;
         $retryThreshold->modify('-' . $retryIntervalMinutes . ' minutes');
 
-        $query->setParameter('retryThreshold', $retryThreshold);
+        $qb = $this->entityManager->createQueryBuilder();
+        $query = $qb->select(array('a'))
```
ShawnTurple, line 122: assuming all this is doing is converting the sql to a Doctrine statement.

ShawnTurple (approved):
Informational: 
Please try to stay within the scope of the ticket, although some of it might not be easy to do because of the troubles you were running into, i suspect some of it could have.  Especially since there is no 'changelog' updates

PR #52, opened 2025-01-14
DESCW-2705 Run database migration automatically during deployment

kshapka-bcgov:
-  Implement a process that automatically triggers database migrations as part of the deployment pipeline.
    - Added a PostSync hook for ArgoCD
- Ensure that the migration process runs smoothly without manual intervention.
    - Changed restart policy to 'never'
    - Added 'no interaction' and 'allow no migration' flags to ensure migrations run without manual intervention.

Screenshot 2025-01-14 at 15 58 48


mhaswell-bcgov:
Have we tried this on the tenant repo? We should be able to test it if we change the dev instance kustomization like this:
```
resources:
- https://github.com/bcgov/naad-connector//deployments/kustomize/overlays/openshift?timeout=120&ref=feature/DESCW-2705-run-database-migration-during-deployment
```
Then point argocd to the branch with that change.

kshapka-bcgov:
> Have we tried this on the tenant repo? We should be able to test it if we change the dev instance kustomization like this:
> 
> ```
> resources:
> - https://github.com/bcgov/naad-connector//deployments/kustomize/overlays/openshift?timeout=120&ref=feature/DESCW-2705-run-database-migration-during-deployment
> ```
> 
> Then point argocd to the branch with that change.

Yes, and I think it's working but I'm trying to spin up a test deployment to ensure sure it works 'out of the box,' so to speak.

PR #54, opened 2025-01-15
Added database-specific ingress network policy

mhaswell-bcgov:


robrien-bcgov (approved):
seems reasonable to restrict ingress to a specific port

ShawnTurple (approved):

PR #56, opened 2025-01-17
DESCW-2847 fetchAlert should not take an array argument.

robrien-bcgov:
- adapted the NaadRepositoryClient test to use fetcHAlert with 2 string args

- refactored the cod that handles missed alerts in the Naad Socket client:
	- passes 'id' and 'sent' strings to the repo client's fetchAlert
	- broke out a function to perform the fetch, validate, filter, and process operations
- refactored the fetchAlert signature and try/catch block to acocunt for the new string params
- updated the catch block to return more granular exception info.

### DIFF
--- a/src/NaadSocketClient.php
+++ b/src/NaadSocketClient.php
@@ -114,7 +114,16 @@ class NaadSocketClient
                 );
 
                 // Fetch, validate, then process missed alerts.
-                $this->handleMissedAlerts($missedAlerts);
+                foreach ( $missedAlerts as $alert ) {
+                    $this->currentOutput = '';
+                    $rawXml = $this->repositoryClient->fetchAlert(
+                        $alert['id'], $alert['sent']
+                    );
+                    $xml = $this->validateResponse($rawXml);
+                    if ($xml) {
+                        $this->processAlert($xml);
+                    }
+                }
             }
         } else {
             $this->processAlert($xml);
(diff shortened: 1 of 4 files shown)


mhaswell-bcgov:
This was interesting because these changes introduced a bug which wasn't being caught by the tests so I added to the tests to catch it (good opportunity to improve the tests). What's happening is because the line to clear the `currentOutput` was removed, every second time `validateAlert()` runs it's adding the alert xml onto the last alert's xml which is making it invalid so it's skipping it and it's not getting inserted. So only half the missed alerts get inserted each time the heartbeat code runs.

src/NaadSocketClient.php
```diff
@@ -140,6 +135,34 @@ public function handleResponse( string $response ): bool
...
+                            $this->repositoryClient->fetchAlert(
+                                $alert['id'], $alert['sent']
+                            )
+                        );
+                    },
+                    $missedAlerts
+                )
+            )
+        );
+    }
```
mhaswell-bcgov, line 164: I find this harder to understand at a glance than the code it's replacing so I'd rather revert it unless there's a good reason not to.

mhaswell-bcgov (changes requested):
Requested changes:

1. Revert creation of handleMissedAlerts() function
2. Fix bug with validation of missed alerts causing them not to get inserted per unit tests (this should be solved by 1).

src/NaadSocketClient.php
robrien-bcgov: I've reverted my changes and just passed the two args to fetchAlert instead. tests are working now too.

mhaswell-bcgov (approved):

PR #57, opened 2025-01-20
DESCW-2771 Adds tests for Entity/Alert class

robrien-bcgov:
[DESCW-2771](https://citz-gdx.atlassian.net/browse/DESCW-2825)

- edits setFailures to handle edge case where $failure is negative.
- tests all trivial setter/getter methods in the same test case
- separates fromXML() Exception and Alert test cases.
image

### DIFF
--- a/src/Entity/Alert.php
+++ b/src/Entity/Alert.php
@@ -148,17 +148,23 @@ class Alert
      *
      * @param int $failures The number of times the alert has
      *                      failed to send to its destination.
+     *                      and set to 0 if negative.
      *
      * @return void
      */
     public function setFailures( int $failures ): void
     {
-        $this->failures = $failures;
+
+        if ($failures < 0) {
+            $this->failures = 0;
+        } else {
+            $this->failures = $failures;
+        }
     }
 
     /**
      * Increments the number of failures.
-     * 
+     *
      * @return void
      */
     public function incrementFailures(): void
@@ -205,7 +211,7 @@ class Alert
             $errorMessage = 'Invalid XML: The "identifier" field is required.';
             throw new Exception($errorMessage);
         }
-        
+
         $alert = new Alert();
         $alert->setId($identifier);
         $alert->setBody($xml->asXML());
--- a/src/Entity/Alert.php
+++ b/src/Entity/Alert.php
@@ -214,10 +214,9 @@ class Alert
         }
 
         $alert = new self();
-        // Avoiding unnecessary calls to getters and setters
-        $alert->id = $identifier;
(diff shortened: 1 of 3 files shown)


tests/AlertTest.php
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /**
+     * Get a valid XML string for testing
+     *
+     * @return string
+     */
+    private function getValidXmlString(): string
+    {
+        return '<?xml version="1.0" encoding="UTF-8"?>
```
mhaswell-bcgov, line 160: We could use `tests/Socket/complete-alert.xml` here which is what a valid alert will look like in production.

src/Entity/Alert.php
```diff
@@ -194,22 +200,24 @@ public function setSuccess( bool $success ): void
...
-        $alert = new Alert();
-        $alert->setId($identifier);
-        $alert->setBody($xml->asXML());
-        $alert->setReceived(new DateTime());
+
+        $alert = new self();
+        // Avoiding unnecessary calls to getters and setters
+        $alert->id = $identifier;
+        $alert->body = $xml->asXML();
+        $alert->received = new DateTime();
```
mhaswell-bcgov, line 220: I think it's slightly better to use the setters here because they have some extra checks in some cases (like setFailures() forcing the value to be >0).

mhaswell-bcgov (approved):

src/Entity/Alert.php
```diff
@@ -194,19 +200,20 @@ public function setSuccess( bool $success ): void
...
+
+        $err = 'Invalid XML: "identifier" field is required and must not be empty.';
+        if ('' === $identifier) {
+            throw new \InvalidArgumentException($err);
         }
-        
-        $alert = new Alert();
+
+        $alert = new self();
         $alert->setId($identifier);
```
robrien-bcgov, line 217: reverted to using setters. this lets us leverage any error-checking logic in the getter/setters.

tests/AlertTest.php
```diff
@@ -0,0 +1,153 @@
...
+     * and body content.
+     *
+     * @return void
+     *
+     * @throws \Exception If there's an error parsing the XML or creating
+     *                    the Alert object.
+     */
+    public function testFromXmlWithValidData(): void
+    {
+        $xmlString = file_get_contents(self::XML_TEST_FILE);
```
robrien-bcgov, line 115: grabs a complete alert XML from the test file

mhaswell-bcgov (approved):

PR #58, opened 2025-01-23
DESCW-2823 bugfix for EntityCollisionException in database.php

robrien-bcgov:
[DESCW-2823 Alerts with msgType of "update" cause database error](https://citz-gdx.atlassian.net/browse/DESCW-2823)

- adds handling for this error
- adds custom logging so its easy to tell which module is throwing

image

### DIFF
--- a/src/Database.php
+++ b/src/Database.php
@@ -23,7 +23,7 @@ use Monolog\Logger;
 class Database
 {
 
-    public EntityManager $entityManager;
+    protected EntityManager $entityManager;
 
     protected Logger $logger;
 
@@ -78,7 +78,6 @@ class Database
      */
     public function insertAlert(Alert $alert): void
     {
-
         // Try to persist the alert, catching duplicate key violations gracefully.
         try {
             $this->entityManager->persist($alert);
@@ -87,7 +86,7 @@ class Database
             $this->logger->error('Unique constraint violation: ' . $e->getMessage());
             $this->entityManager = $this->getEntityManager();
         } catch (EntityIdentityCollisionException $e) {
-            $this->logger->critical(
+            $this->logger->error(
                 'Entity identity collision: ' . $e->getMessage()
             );
             $this->entityManager->detach($alert);
--- a/src/Database.php
+++ b/src/Database.php
@@ -6,8 +6,10 @@ use Bcgov\NaadConnector\Entity\Alert;
 use Bcgov\NaadConnector\NaadVars;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
+use Doctrine\ORM\Exception\EntityIdentityCollisionException;
 use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\ORMSetup;
+use Monolog\Logger;
 
 /**
  * Database class for interacting with the database.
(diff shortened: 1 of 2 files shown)


src/Database.php
```diff
@@ -69,13 +76,21 @@ protected function getEntityManager(): EntityManager
...
             $this->flush();
-        } catch ( UniqueConstraintViolationException $e ) {
+        } catch (UniqueConstraintViolationException $e) {
+            $this->logger->error('Unique constraint violation: ' . $e->getMessage());
             $this->entityManager = $this->getEntityManager();
+        } catch (EntityIdentityCollisionException $e) {
+            $this->logger->critical(
+                'Entity identity collision: ' . $e->getMessage()
+            );
+            $this->entityManager->detach($alert);
```
robrien-bcgov, line 93: The line `$this->entityManager->detach($alert);` within the `insertAlert` method is used to detach the given `$alert` entity from the `EntityManager`. This means that the `EntityManager` will no longer manage the `$alert` entity, and any changes made to it will not be tracked or persisted to the database. This is typically done to prevent further operations on an entity that has caused an error, such as an `EntityIdentityCollisionException`, ensuring it does not interfere with subsequent database operations.

src/Database.php
```diff
@@ -69,13 +76,21 @@ protected function getEntityManager(): EntityManager
...
     {
+
+        // Try to persist the alert, catching duplicate key violations gracefully.
         try {
             $this->entityManager->persist($alert);
             $this->flush();
-        } catch ( UniqueConstraintViolationException $e ) {
+        } catch (UniqueConstraintViolationException $e) {
+            $this->logger->error('Unique constraint violation: ' . $e->getMessage());
             $this->entityManager = $this->getEntityManager();
```
robrien-bcgov, line 88: The line `$this->entityManager = $this->getEntityManager();` in the `insertAlert` method reinitializes the `EntityManager` after catching a `UniqueConstraintViolationException`. This is done to reset the state of the `EntityManager` and ensure it is in a consistent state for subsequent operations, especially after encountering an error that might have left it in an invalid state.

src/Database.php
mhaswell-bcgov: How can I reproduce this exception? I can get the other one to happen but not sure about this one after playing around with it for a bit.

src/Database.php
```diff
@@ -21,14 +23,19 @@
 class Database
 {
 
-    protected EntityManager $entityManager;
+    public EntityManager $entityManager;
```
mhaswell-bcgov, line 26: This should still be protected.

src/Database.php
```diff
@@ -69,13 +76,21 @@ protected function getEntityManager(): EntityManager
...
+        // Try to persist the alert, catching duplicate key violations gracefully.
         try {
             $this->entityManager->persist($alert);
             $this->flush();
-        } catch ( UniqueConstraintViolationException $e ) {
+        } catch (UniqueConstraintViolationException $e) {
+            $this->logger->error('Unique constraint violation: ' . $e->getMessage());
             $this->entityManager = $this->getEntityManager();
+        } catch (EntityIdentityCollisionException $e) {
+            $this->logger->critical(
```
mhaswell-bcgov, line 90: I've been using `critical` logs only for situations that cause the application to exit so this would be an `error` since it can recover from it, at least the way I've been doing the log levels so far.

mhaswell-bcgov (changes requested):

src/Database.php
robrien-bcgov: just add this code to the function above the try/catch block in insertAlert()

```php
            $collisionId = 'urn:oid:2.49.0.1.124.0745642605.2024';
            $alert->setId($collisionId);

            // Find an existing entity with the same ID to ensure it's managed
            $existingAlert = $this->entityManager->find(Alert::class, $collisionId);
            if (!$existingAlert) {
                // If no existing entity is found, create and persist one to simulate the collision
                $existingAlert = new Alert();
                $existingAlert->setId($collisionId);
                $existingAlert->setBody('<alert>existing entity</alert>');
                $existingAlert->setReceived(new \DateTime());
                $this->entityManager->persist($existingAlert);
                $this->flush();
            }
            ```
            
            
            
image

robrien-bcgov: it fires off when the client fetches from the NAAD repository

mhaswell-bcgov (approved):

PR #60, opened 2025-02-05
[DESCW-2167] Improved logging across application

mhaswell-bcgov:
Refactored CustomLogger class into LoggerFactory class as it was extending final class Monolog\Logger
Added LOG_LEVEL environment variable to devcontainer config, deployments, application
Changed NAAD repo url to use https
Improved log messages and log levels across codebase
Added logger to NaadRepositoryClient
Updated unit tests

### DIFF
--- a/src/LoggerFactory.php
+++ b/src/LoggerFactory.php
@@ -6,7 +6,7 @@ use Monolog\Handler\StreamHandler;
 use Monolog\Processor\PsrLogMessageProcessor;
 
 /**
- * A class for setting up a custom Monolog logger instance
+ * A class for creating Monolog logger instances
  * with configurable channel name and logging level.
  *
  * @category Logging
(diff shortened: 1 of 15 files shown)


src/NaadVars.php
```diff
@@ -82,7 +85,8 @@ class NaadVars
     private static array $defaultValues = [
         'naadName'     => 'NAAD-1',
         'naadUrl'      => 'streaming1.naad-adna.pelmorex.com',
-        'naadRepoUrl'  => 'capcp1.naad-adna.pelmorex.com'
+        'naadRepoUrl'  => 'capcp1.naad-adna.pelmorex.com',
+        'logLevel'     => 'info',
```
ShawnTurple, line 89: would this be the default level for production?

ShawnTurple (approved):

src/NaadVars.php
mhaswell-bcgov: Yes. I think it's useful to have the info logs (heartbeats are debug level now) on production, otherwise there will be no logs at all unless an error occurs.

src/LoggerFactory.php
```diff
@@ -39,7 +37,7 @@ public function __construct(
             new StreamHandler('php://stdout', $logLevel),
         ];
```
robrien-bcgov, line 38: nice & simple

robrien-bcgov (approved):

PR #61, opened 2025-02-10
[DESCW-2317] Writing logs to files and log rotation

mhaswell-bcgov:


.devcontainer/devcontainer.env
```diff
@@ -19,4 +19,6 @@ NAAD_PORT=8080
 NAAD_NAME="NAAD-1"
 NAAD_URL="streaming1.naad-adna.pelmorex.com"
 NAAD_REPO_URL="capcp1.naad-adna.pelmorex.com"
-LOG_LEVEL=debug
\ No newline at end of file
+LOG_LEVEL=debug
+LOG_PATH=/logs/naad.log
+LOG_RETENTION_DAYS=7
```
ShawnTurple, line 24: in production this should be a minimum of 30 days

deployments/kustomize/base/app/kustomization.yaml
```diff
@@ -4,6 +4,7 @@ resources:
 - deployment.yaml
 - deployment2.yaml
 - job.yaml
+- volume.yaml
 
 # Updates the labels for all containers.
 commonLabels:
```
robrien-bcgov, line 10: i keep getting warnings that 'commonLabels' is deprecated and should be replaced with 'labels' instead

src/LoggerFactory.php
```diff
@@ -35,6 +41,11 @@ public static function createLogger(
         $logLevel = self::_convertLogLevel($level);
         $handlers = [
             new StreamHandler('php://stdout', $logLevel),
+            new RotatingFileHandler(
```
ShawnTurple, line 44: do these handlers run in parallel?

ShawnTurple (approved):

deployments/kustomize/base/app/kustomization.yaml
mhaswell-bcgov: Yeah I got those too. We can update them automatically with a kustomize command I think but I'll leave it for now.

src/LoggerFactory.php
mhaswell-bcgov: Yes the writes go through both handlers.

PR #63, opened 2025-02-11
Feature/descw 2315 clean up old alerts

robrien-bcgov:
once per week, this Cronjob will delete all but the 100 freshest alerts, with logging

https://citz-gdx.atlassian.net/browse/DESCW-2315

### DIFF
--- a/deployments/kustomize/base/app/alert-cleanup.yaml
+++ b/deployments/kustomize/base/app/alert-cleanup.yaml
@@ -1,0 +1,42 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: alert-cleanup
+spec:
+  schedule: "0 0 * * 0"  # Runs once a week on Sunday at midnight.
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: OnFailure
+          volumes:
+            - name: logs-volume
+              persistentVolumeClaim:
+                claimName: naad-logs
+          containers:
+          - name: alert-cleanup
+            image: bcgovgdx/naad-app
+            imagePullPolicy: IfNotPresent
+            envFrom:
+            - configMapRef:
+                name: cron-alert-cleanup-config
+            - configMapRef:
+                name: naad-shared-config
+            - configMapRef:
+                name: database-config
+            - secretRef:
+                name: naad-secrets
+            - secretRef:
+                name: database-secrets
+            volumeMounts:
+            - name: logs-volume
+              mountPath: /var/log/app
+              subPath: cron
+            resources:
+              requests:
+                cpu: 100m
+                memory: 256Mi
+              limits:
+                cpu: 500m
+                memory: 512Mi
+            command: ["/bin/sh", "-c", "php src/alert-cleanup.php"]
--- a/src/NaadVars.php
+++ b/src/NaadVars.php
@@ -41,8 +41,10 @@ namespace Bcgov\NaadConnector;
  * @property-read string $naadRepoUrl The URL of the NAAD Repository, which we
  * fetch missing alerts from, eg. "capcp1.naad-adna.pelmorex.com".
  * @property-read string $logLevel The minimum level of logs the Logger will send.
+ * before rotating.
+ * @property-read int $alertsToKeep The number of alerts to keep in the database.
  * @property-read string $logPath The path to the log file to write to.
- * @property-read int    $logRetentionDays The number of days to keep a log file 
+ * @property-read int    $logRetentionDays The number of days to keep a log file
  * before rotating.
  */
 class NaadVars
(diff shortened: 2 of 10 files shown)


src/alert-cleanup.php
```diff
@@ -0,0 +1,31 @@
...
+);
+$alertCleanupLogger         = $logger->withName('AlertCleanup');
+
+// Create a new Database Instance so we can use it to delete old alerts.
+$database = new Database($alertCleanupLogger);
+
+if (method_exists($database, 'deleteOldAlerts')) {
+    $database->deleteOldAlerts();
+} else {
+    $alertCleanupLogger->error(
```
mhaswell-bcgov, line 28: I don't think this is necessary. `deleteOldAlerts()` must exist if this code does.

src/alert-cleanup.php
```diff
@@ -0,0 +1,31 @@
+<?php
+require_once 'vendor/autoload.php';
+
+// Require headers.php from /src directory
+$headers = include dirname(__DIR__) . '/src/headers.php';
```
mhaswell-bcgov, line 5: This is unused as far as I can tell.

src/NaadVars.php
```diff
@@ -41,8 +41,12 @@
  * @property-read string $naadRepoUrl The URL of the NAAD Repository, which we
  * fetch missing alerts from, eg. "capcp1.naad-adna.pelmorex.com".
  * @property-read string $logLevel The minimum level of logs the Logger will send.
+ * @property-read int $alertsToKeep The number of alerts to keep in the database.
+ * @property-read int    $logRetentionDays The number of days to keep a log file
+ * before rotating.
+ * @property-read int $alertsToKeep The number of alerts to keep in the database.
  * @property-read string $logPath The path to the log file to write to.
- * @property-read int    $logRetentionDays The number of days to keep a log file 
+ * @property-read int    $logRetentionDays The number of days to keep a log file
```
mhaswell-bcgov, line 49: $logRetentionDays and $alertsToKeep are duplicated here.

deployments/kustomize/base/app/alert-cleanup.yaml
```diff
@@ -0,0 +1,42 @@
...
+            - configMapRef:
+                name: database-config
+            - secretRef:
+                name: naad-secrets
+            - secretRef:
+                name: database-secrets
+            volumeMounts:
+            - name: logs-volume
+              mountPath: /var/log/app/
+              subPath: naad-1
```
mhaswell-bcgov, line 34:
```suggestion
              subPath: cron
```

I think we should separate the actual naad-1 application logs from the cron job logs. This will also require another configmap, something like:

```
  - name: cron-config
    literals:
    - NAAD_NAME=NAAD-CRON
    - NAAD_URL=streaming1.naad-adna.pelmorex.com # Not actually necessary since it doesn't connect to the socket but I think the application will complain if we don't include something here
    - NAAD_REPO_URL=capcp1.naad-adna.pelmorex.com # Same as above
    - LOG_PATH=/var/log/app/cron/app.log
```

mhaswell-bcgov (changes requested):

deployments/kustomize/base/app/alert-cleanup.yaml
mhaswell-bcgov: It shouldn't since it's a mounted volume. That subPath will be unmounted but the log files it writes to should stay on the volume.

src/alert-cleanup.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    $database = new Database($logger);
+    $database->deleteOldAlerts();
+}
+
+// Call the function to delete old alerts
+deleteOldAlerts($logger);
+
+// Pause for 120 seconds to allow the deletion process to complete
+sleep(120);
```
robrien-bcgov, line 37: bonus: keeps the pod alive long enough to verify the volume subpath has a new alert log written to it.

src/alert-cleanup.php
```diff
@@ -1,31 +1,37 @@
...
+ * Deletes old alerts from the database.
+ *
+ * This function connects to the database and removes alerts that are considered old
+ * based on predefined criteria. It uses the provided logger to log the process.
+ *
+ * @param Logger $logger The logger instance used for logging the deletion process.
+ *
+ * @return void
+ */
+function deleteOldAlerts(Logger $logger)
```
mhaswell-bcgov, line 27: Why do we need this function?

src/alert-cleanup.php
mhaswell-bcgov: Useful for testing but once we confirm it writes to the volume I don't think we need it.

mhaswell-bcgov (dismissed):

mhaswell-bcgov (approved):

PR #64, opened 2025-02-13
DESCW-2867 fixes deprecated 'CommonLabels' error on deploys

robrien-bcgov:
Optional quality-of-life improvement for when we deploy.

### DIFF
--- a/deployments/kustomize/base/app/kustomization.yaml
+++ b/deployments/kustomize/base/app/kustomization.yaml
@@ -26,13 +26,6 @@ configMapGenerator:
     - NAAD_URL=streaming1.naad-adna.pelmorex.com
     - NAAD_REPO_URL=capcp1.naad-adna.pelmorex.com
     - LOG_PATH=/var/log/app/cron/alert-cleanup.log
-    - DESTINATION_URL="http://local:38080/embc/wp-json/naad/v1/alert"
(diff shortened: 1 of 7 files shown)


deployments/kustomize/base/app/kustomization.yaml
```diff
@@ -7,39 +7,41 @@ resources:
...
-    type: Opaque
-    literals:
-    - DESTINATION_PASSWORD='AAAA AAAA AAAA AAAA'
\ No newline at end of file
+- name: naad-secrets
+  literals:
+    - DESTINATION_PASSWORD='AAAA AAAA AAAA AAAA'
+  type: Opaque
+
+labels:
```
mhaswell-bcgov, line 43: Thanks for fixing these. Did `kustomize edit fix` do this automatically?

mhaswell-bcgov (approved):

deployments/kustomize/base/app/kustomization.yaml
robrien-bcgov: yes, i had to cd to the directory where the files were, and run `kustomize edit fix`, and it mostly worked. I think the problem was it changed the format of the file in a way I didn't like ( the -name got moved ), but otherwise it worked great!

PR #65, opened 2025-02-14
[DESCW-2824] Added failure threshold to DestinationClient so that the application …

mhaswell-bcgov:
…crashes when an alert fails more than 5 times

Added unit test case for the above

### DIFF
--- a/tests/DestinationClientTest.php
+++ b/tests/DestinationClientTest.php
@@ -146,6 +146,40 @@ final class DestinationClientTest extends TestCase
     }
 
     /**
+     * Tests sendAlerts method when and alert request fails and its
+     * failure threshold is reached.
+     *
+     * @return void
+     */
+    #[Test]
+    public function testSendAlertsFailureThresholdReached()
+    {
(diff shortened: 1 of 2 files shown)


tests/DestinationClientTest.php
```diff
@@ -145,6 +145,40 @@ public function testSendAlertsFailure()
...
+    /**
+     * Tests sendAlerts method when and alert request fails and its
+     * failure threshold is reached.
+     *
+     * @return void
+     */
+    #[Test]
+    public function testSendAlertsFailureThresholdReached()
+    {
+        $alert = $this->createMock(Alert::class);
```
robrien-bcgov, line 157: is this basically telling the mock to return '5'?
or is it predicting that it will?

robrien-bcgov (approved):

tests/DestinationClientTest.php
mhaswell-bcgov: It's forcing it to return 5 so that it will trigger the code that checks for that threshold.

PR #68, opened 2025-02-18
Bug/descw 2863 liveness probe inactive openshift

robrien-bcgov:
descw-2863 fix liveness probe that wasn't firing off
https://citz-gdx.atlassian.net/browse/DESCW-2863

- The trouble was that the liveness command wasn't giving the exit status.
it was just echoing a 1 or 0.


.devcontainer/sample.override.env
```diff
@@ -5,6 +5,6 @@
 MARIADB_ROOT_PASSWORD=rootpassword
 
 # App environment vars
-DESTINATION_URL="http://local:38080/sitename/wp-json/naad/v1/alert"
+DESTINATION_URL="http://0.0.0.0:38080/sitename/wp-json/naad/v1/alert"
```
mhaswell-bcgov, line 8: Does this URL still work locally?

liveness.sh
```diff
@@ -1,12 +1,19 @@
...
+if [ -f "$HEARTBEAT_FILE" ]; then
+    # Get the last modified time of the file in seconds since epoch
+    LAST_MODIFIED=$(stat -c %Y "$HEARTBEAT_FILE")
+    # Get the current time in seconds since epoch
+    CURRENT=$(date +%s)
+
+    # If the file was modified in the last 120 seconds, exit with success
+    if [ $((CURRENT - LAST_MODIFIED)) -le 120 ]; then
+        exit 0
+    fi
```
mhaswell-bcgov, line 15: Have we tried this on OpenShift? I'm guessing the old script used commands that don't exist on OpenShift (`date`)? Just trying to understand what caused it not to work before.

mhaswell-bcgov (commented):
Looking for more info before I approve.

.devcontainer/sample.override.env
robrien-bcgov: on my machine it halfway works (in dev container mode)


What I mean is that I don't get a 404; It just gives me a 'no route' error:

image

I probably just have to set up my naad plugin differently or something.

liveness.sh
robrien-bcgov: yes. 

It's not pretty, but it will restart the container until it hits a crashloop backoff state.

My theory here is that it wasn't working before because the `liveness probe` expects a `return value` of 0 or 1 from the script.

Before, the script was not returning a return value, it was echoing 0 or 1. 

My guess is that when a script echos '1' or '0', then exits, it has a return value of 1.

since that's success, the liveness probe never has a reason to fire off.

deployments/kustomize/base/app/kustomization.yaml
```diff
@@ -28,7 +28,7 @@ configMapGenerator:
     - LOG_PATH=/var/log/app/cron/alert-cleanup.log
 - name: naad-shared-config
   literals:
-    - DESTINATION_URL="http://local:38080/embc/wp-json/naad/v1/alert"
+    - DESTINATION_URL="http://0.0.0.0:38080/embc/wp-json/naad/v1/alert"
```
robrien-bcgov, line 31: I think `http://local:38080...` was a typo or something?

that hostname doesn't seem legit to me

.devcontainer/sample.override.env
robrien-bcgov: Also, the tests work with that URL, (not that that is the best evidence)

deployments/kustomize/base/app/kustomization.yaml
mhaswell-bcgov: It's `local` because of this: https://github.com/bcgov/naad-connector/blob/main/.devcontainer/docker-compose.yml#L23
`localhost` or `0.0.0.0` when the code is executing inside a container is that container's `localhost`, but we need it to point to the host machine's `localhost` which is where the WP site is accessible. In this case (outside of the .devcontainer directory) it doesn't really matter what we put here because it won't work. It needs to point to a WP site somewhere.

liveness.sh
robrien-bcgov: * exit code, not return value

mhaswell-bcgov (approved):

PR #69, opened 2025-02-19
DESCW-2921 Secondary fix to prevent crashloops

kshapka-bcgov:
This pull request adds initContainers to the job and both socket deployments.

### Why
The database migration and sockets could crash in the case that they attempt to edit the database before it is ready to accept connections. This happens infrequently for the sockets and regularly for the migration job.

### How
An initContainer is spun up before primary pod deployment, which runs until it determines that the database port is open. After this, the initContainer exits and the primary pod is deployed.

### Considerations
This slows the deployment somewhat, as two pods need to initialize for each component.

### DIFF
--- a/deployments/kustomize/base/app/deployment.yaml
+++ b/deployments/kustomize/base/app/deployment.yaml
@@ -17,7 +17,7 @@ spec:
       containers:
         - name: naad-app
           image: bcgovgdx/naad-app
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: Always
           envFrom:
           - configMapRef:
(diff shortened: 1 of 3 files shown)


mhaswell-bcgov:
I think this is ok if you want to move it out of draft. The more we change these files the more I feel like we need to figure out a way to do less code duplicating but that's a problem for another day.

kshapka-bcgov:
@mhaswell-bcgov Completely agree -Working on these has made me think that it would be better to slim down at least the socket deployments.

mhaswell-bcgov (approved):

PR #70, opened 2025-02-20
Fix permission issue that was preventing heartbeat.log update and triggering liveness probe

robrien-bcgov:
I changed the permissions on heartbeat.log in the Dockerfile to 664, thinking that was a good idea for security (Owner and Group can read/write, but others can only read)

However, on Openshift, the 'others' could no longer write to the log, and the error showed a permissions issue, which stopped the heartbeat from being updated.

That caused the liveness probe to trigger, which meant a restart.

That became a crash loop.

I've fixed the permissions to be more open, and rebuilt the image in the tools namespace, which seems to have fixed it on openshift.

This PR will consolidate those changes.


mhaswell-bcgov (approved):
Thanks for following up on this and fixing it.

PR #74, opened 2025-02-28
DESCW-2821 Openshift Image Build Pipeline

kshapka-bcgov:


kshapka-bcgov:
@mhaswell-bcgov Can you please confirm the secrets if they make sense to you?

mhaswell-bcgov (approved):
This is outside of my wheelhouse but it looks reasonable. I guess we're not able to run the actual GH workflow until it's merged in?

kshapka-bcgov:
Annoyingly I don't think there is a way to test the workflow itself without it being in main.

PR #75, opened 2025-02-28
DESCW-2764 fix broken oc command in workflow

kshapka-bcgov:


mhaswell-bcgov (approved):
Nice, saw the workflow passed and I'm seeing a new `dev` image.

PR #76, opened 2025-03-01
Feature/descw 2753 integrate vault secrets

robrien-bcgov:
This is a work in progress.

mostly it is here to track my changes while testing the vault secrets across the database, naad-1, naad-2, database migrate job, and likely the cronjob for deleting alerts as well since that needs database secrets.


robrien-bcgov:
this was just for testing

PR #77, opened 2025-03-05
DESCW-2920 Expand response handling for non-fatal errors.

kshapka-bcgov:


robrien-bcgov (approved):

src/DestinationClient.php
```diff
@@ -116,6 +109,46 @@ public function sendAlerts(): bool
...
+     * Log an HTTP response to a failed attempt to send an alert.
+     *
+     * @param Alert          $alert    The Alert in the failed send attempt.
+     * @param array|null     $response The HTTP response from WordPress, if any.
+     * @param Exception|null $e        The provided exception, if any.
+     *
+     * @return void
+     */
+    private function handleAlertFailure(
+        $alert, array $response = null, \Exception $e = null
```
mhaswell-bcgov, line 122: Looks like the `$response` argument is always given when this function is called so it doesn't need to be optional.

src/DestinationClient.php
```diff
@@ -76,28 +77,20 @@ public function sendAlerts(): bool
...
                     );
+                } elseif (500 <= $response['status_code']
+                    || in_array($response['status_code'], self::FATAL_ERRORS, true)
+                ) {
+                    // 5xx and certain 4xx errors are cause for a retry.
+                    throw new \Exception('Unexpected status code received.');
+
                 } else {
-                    throw new \Exception('Non-200 status code recieved.');
+                    // All other response codes are unrecoverable.
```
mhaswell-bcgov, line 87: Are these comments swapped? 5xx and some 4xx errors are fatal/unrecoverable and the rest can be retried. The logic in this function is complicated since before this PR so kind of hard to follow.

mhaswell-bcgov (approved):

src/DestinationClient.php
kshapka-bcgov: This is to handle the old logic where $response might not be set.

src/DestinationClient.php
kshapka-bcgov: Oh, it should probably be 'cause for intervention.' I wrote these before we spoke last.

PR #78, opened 2025-03-10
DESCW-2949 Adding vault secret variable logic

ShawnTurple:


src/NaadVars.php
```diff
@@ -116,8 +116,13 @@ public function __construct()
         foreach (self::$envMap as $property => $envKey) {
             // Get the environment variable out of the .env file.
             // If not found there, use a default, if not there, just be null.
-            $this->$property = getenv($envKey)
+            $secretFileName = sprintf("/vault/secrets/%s", $envKey);
+            if (file_exists($secretFileName) ) {
+                $this->$property = rtrim(file_get_contents($secretFileName), "\r\n");
```
robrien-bcgov, line 121: does that mean I need to change the annotation so that a new file will be created for each vault secret?

robrien-bcgov (approved):
seems like this will make secrets easier to work with.

mhaswell-bcgov (approved):

mhaswell-bcgov (approved):

PR #79, opened 2025-03-12
Feature/descw 2948 update kustomize structure

ShawnTurple:


deployments/kustomize/base2/database-migration-job.yaml
```diff
@@ -0,0 +1,37 @@
...
+    app.kubernetes.io/name: database-migration
+    argocd.argoproj.io/hook: PostSync
+    argocd.argoproj.io/hook-delete-policy: HookSucceeded
+spec:
+  template:
+    spec:
+      restartPolicy: OnFailure
+      containers:
+        - name: naad-database-migration
+          image: bcgovgdx/naad-app
```
mhaswell-bcgov, line 15:
```suggestion
          image: image-registry.openshift-image-registry.svc:5000/bdf349-tools/naad-app
```
Should match the image used in the stateful set I think. I assume it was already pointing to the wrong image before any of these changes.

deployments/kustomize/README.md
```diff
@@ -0,0 +1,107 @@
...
+# Config
+
+# default values for nad url and repo url, however the app will have logic to overwrite this info.
+NAAD_URL=streaming1.naad-adna.pelmorex.com # will become deprecated
+NAAD_REPO_URL=capcp1.naad-adna.pelmorex.com # will become deprecated 
+# The path where the logs get stored, each log instance will create a subpath and the corresponding log files.
+LOG_PATH=/var/log/app/naad/app.log
+# info
+LOG_LEVEL=info
+# 0 - no retention, > 0 number of days
```
mhaswell-bcgov, line 43:
```suggestion
# 0 - unlimited retention, > 0 number of days
```

deployments/kustomize/base2/app-statefulset.yaml
```diff
@@ -0,0 +1,67 @@
...
+        - name: logs-volume
+          persistentVolumeClaim:
+            claimName: naad-logs
+      terminationGracePeriodSeconds: 10
+      containers:
+        - name: naad-app
+          image: image-registry.openshift-image-registry.svc:5000/bdf349-tools/naad-app
+          imagePullPolicy: Always
+          env:
+          - name: POD_INDEX
```
mhaswell-bcgov, line 31: I think there's a better name for this. The application isn't necessarily being run in a pod so a more environment-agnostic name might be something like `FEED_INDEX` or `FEED_ID` (NAADS refers to its own sockets as feeds in their documentation).

deployments/kustomize/README.md
```diff
@@ -0,0 +1,107 @@
+# Deployments
+The OpenShift deployment for Early Earthquake System (NAAD) which uses [kustomize](https://kubectl.docs.kubernetes.io/guides/introduction/kustomize/), and all commands are used from the root of this repository.
```
mhaswell-bcgov, line 2:
```suggestion
The OpenShift deployment for Early Earthquake Warning System (NaadConnector) which uses [kustomize](https://kubectl.docs.kubernetes.io/guides/introduction/kustomize/), and all commands are used from the root of this repository.
```

README.md
```diff
@@ -110,7 +111,7 @@ oc start-build naad-app --follow
 4. paste the login command into your terminal.
 5. switch to the correct (12345-dev) namespace.
 6. (OPTIONAL - for testing) edit your kustomization file to replace the value of `resources:` to:
-    - `https://github.com/bcgov/naad-connector//deployments/kustomize/overlays/openshift?ref=<your-feature-branch>`
+    - `https://github.com/bcgov/naad-connector/deployments/kustomize/overlays/openshift?ref=<your-feature-branch>`
```
mhaswell-bcgov, line 114: Correct me if I'm wrong but I think the double `/` is actually required because the part before the `//` is a url and the part after is a file path inside the repo, not a url that actually exists.

mhaswell-bcgov (changes requested):
Might have been easier to review if we just made the changes directly to `base/` although I understand that it would break current deployments. Maybe this should merge into a larger feature branch where we can also merge the required changes to the source code before going into `main`.

deployments/kustomize/base2/app-statefulset.yaml
ShawnTurple: ok i will update it to `FEED_ID`

deployments/kustomize/base2/database-migration-job.yaml
ShawnTurple: actually the stateful set needs to be updated, we shouldn't have the final image here, we should let kustomize update this, and if `bcgovgdx/naad-app` then it is easier

deployments/kustomize/README.md
ShawnTurple: honestly i don't know i thought it was an error, but i will add it back in, it has no affect on functionality.

PR #81, opened 2025-03-13
[DESCW-2933] Prevent sending duplicate alerts with same id

mhaswell-bcgov:


robrien-bcgov (approved):

ShawnTurple (approved):
we might want to consider the recommendations by sonarcube wit the complexity of the function if/else nesting

PR #85, opened 2025-03-27
[DESCW-2966] Fix bug in Database::deleteOldAlerts() causing out of memory errors

mhaswell-bcgov:


src/Database.php
```diff
@@ -191,4 +161,45 @@ public function getUnsentAlerts(): array
...
+
+        // Get the current total number of rows in the table.
+        $qb = $this->entityManager->createQueryBuilder();
+        $count = $qb->select('count(a.id)')
+            ->from(Alert::class, 'a')
+            ->getQuery()
+            ->getSingleScalarResult();
+
+        // Calculate the number of rows to delete.
+        $alertsToDelete = intval($count) - $alertsToKeep;
```
ShawnTurple, line 183: to me this should be passed in as a parameter, than this function might be more testable.

ShawnTurple (dismissed):
are there any tests for this?

src/Database.php
mhaswell-bcgov: I agree, moved `$alertsToKeep` into an argument which is passed in from the alert-cleanup script.

mhaswell-bcgov:
> are there any tests for this?

@ShawnTurple We have no tests for this class because it requires a database connection. Would probably be worth looking into how to write tests for it.

robrien-bcgov (approved):
LGTM :pear:

ShawnTurple (approved):

PR #86, opened 2025-03-27
DESCW-2947 Remove init containers for migrations and socket deployments

kshapka-bcgov:
It does not seem to be possible to alter the database class itself to enact the requested changes, since the doctrine migration script has no retry option and immediately exits on migration failure.

I settled on creating my own script which could be wrapped in retries.

---
This **won't work** for socket deployments, but it is relatively rare for sockets to complete deployment before the database service is ready, so the pods restarting in this case might be useful as visual indicators that something isn't quite right.

---
### Sample migration log
```
 Migrating database (Attempt: 1/5)...

In ExceptionConverter.php line 80:
                                                                               
  An exception occurred in the driver: SQLSTATE[HY000] [2002] Connection time  
  d out                                                                        
                                                                               

In Exception.php line 28:
                                               
  SQLSTATE[HY000] [2002] Connection timed out  
                                               

In PDOConnect.php line 23:
                                               
  SQLSTATE[HY000] [2002] Connection timed out  
                                               

migrations:migrate [--write-sql [WRITE-SQL]] [--dry-run] [--query-time] [--allow-no-migration] [--all-or-nothing [ALL-OR-NOTHING]] [--no-all-or-nothing] [--configuration CONFIGURATION] [--em EM] [--conn CONN] [--db-configuration DB-CONFIGURATION] [--] [<version>]

Migration failed. Retrying in 10 seconds...
Migrating database (Attempt: 2/5)...

In ExceptionConverter.php line 80:
                                                                               
  An exception occurred in the driver: SQLSTATE[HY000] [2002] Connection refu  
  sed                                                                          
                                                                               

In Exception.php line 28:
                                             
  SQLSTATE[HY000] [2002] Connection refused  
                                             

In PDOConnect.php line 23:
                                             
  SQLSTATE[HY000] [2002] Connection refused  
                                             

migrations:migrate [--write-sql [WRITE-SQL]] [--dry-run] [--query-time] [--allow-no-migration] [--all-or-nothing [ALL-OR-NOTHING]] [--no-all-or-nothing] [--configuration CONFIGURATION] [--em EM] [--conn CONN] [--db-configuration DB-CONFIGURATION] [--] [<version>]

Migration failed. Retrying in 10 seconds...
Migrating database (Attempt: 3/5)...
[notice] Migrating up to NaadConnector\Migrations\Version20241004171841
[notice] finished in 30.4ms, used 6M memory, 2 migrations executed, 2 sql queries

 [OK] Successfully migrated to version:                                         
      NaadConnector\Migrations\Version20241004171841                            
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,12 +18,13 @@ COPY --from=composer:latest /usr/bin/composer /usr/local/bin/composer
 # Copy application source code (.dockerignore masks out all the unneeded files from the root directory)
 COPY ./ ./
 
-# Copy entrypoint script into /app and set permissions
+# Copy entrypoint and migration scripts into /app
 COPY ./entrypoint.sh /app/entrypoint.sh
-RUN chmod +x /app/entrypoint.sh /app/src/alert-cleanup.php
+COPY ./run-migrations.sh /app/run-migrations.sh
 
 # Create non-root user, set permissions, and prepare the environment
-RUN useradd -m -u 1001 appuser && \
+RUN chmod +x /app/entrypoint.sh /app/src/alert-cleanup.php /app/run-migrations.sh && \
+    useradd -m -u 1001 appuser && \
     mkdir -p /app/vendor && \
     touch /app/heartbeat.log && \
     chown -R 1001:0 /app && \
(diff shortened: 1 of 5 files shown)


mhaswell-bcgov (commented):
Can we remove this too now? https://github.com/bcgov/naad-connector/blob/main/deployments/kustomize/base/kustomization.yaml#L31 or will this mess something up with the deployments?
Not sure if there are other places where these env variables are referenced, maybe also in the tenant repo? @ShawnTurple

mhaswell-bcgov (approved):

PR #87, opened 2025-03-31
DESCW-2964 Fix connection failSpam bug with short delay between retries.

robrien-bcgov:
https://citz-gdx.atlassian.net/browse/DESCW-2964

### DIFF
--- a/src/NaadSocketConnection.php
+++ b/src/NaadSocketConnection.php
@@ -80,7 +80,7 @@ class NaadSocketConnection
                 $this->logger->info(
                     'Socket connected. Listening for socket messages...'
                 );
-    
+
                 $this->setEventHandlers($connection);
             },
             // Unsuccessful connection, get an Exception.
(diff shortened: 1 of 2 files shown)


mhaswell-bcgov (approved):

src/NaadSocketConnection.php
```diff
@@ -100,16 +100,16 @@ function (Exception $e) {
...
      * @return void
      * @link   https://github.com/reactphp/stream?tab=readme-ov-file#readablestreaminterface
      */
     protected function setEventHandlers(ConnectionInterface $connection)
     {
-        $this->reconnectAttempts = 0;
 
         $connection->on(
             'data', function (string $chunk) {
+                $this->reconnectAttempts = 0;
```
kshapka-bcgov, line 112: Looks like you're most of the way to fixing the other bug too. :)

robrien-bcgov: Yeah i might just mark it done

PR #88, opened 2025-04-16
Feature/notify 106

kshapka-bcgov:


ShawnTurple (approved):
Everything looks great here, the only think that might seem out of place is the placement of the sidecar.yaml, as it is not part of the base kustomize deployments, so i worry about where it is located. some thoughts might be `deployments/utilities/sidecar.yaml` or `deployments/sidecar.yaml` - No changes required

mhaswell-bcgov:
Although we'll likely need something like this for naad-connector too this ticket was meant to add the pod to [des-notifybc](https://github.com/bcgov/des-notifybc).

PR #91, opened 2025-04-29
Feature/descw 2987

kshapka-bcgov:
- Adds custom exception classes
- Adds unit tests for exception classes and modifies existing tests to use them
- ~~Small fix to Alert.php naming convention: $send_attempted -> $sendAttempted.~~

### DIFF
--- a/src/Entity/Alert.php
+++ b/src/Entity/Alert.php
@@ -34,7 +34,7 @@ class Alert
     private ?DateTime $received = null;
 
     #[ORM\Column]
-    private ?DateTime $sendAttempted = null;
+    private ?DateTime $send_attempted = null;
 
     #[ORM\Column]
     private int $failures = 0;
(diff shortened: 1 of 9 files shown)


src/Entity/Alert.php
```diff
@@ -34,7 +34,7 @@ class Alert
     private ?DateTime $received = null;
 
     #[ORM\Column]
-    private ?DateTime $send_attempted = null;
+    private ?DateTime $sendAttempted = null;
```
mhaswell-bcgov, line 37: Getting an error:
```
Fatal error: Uncaught PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 't0.sendAttempted' in 'SELECT' in /workspaces/naad-connector/vendor/doctrine/dbal/src/Driver/PDO/Statement.php:55
```
I know Sonar complains about it but I don't know how to make this change since the variable name needs to match the column name in the database.

mhaswell-bcgov (changes requested):

mhaswell-bcgov (approved):
Looks good now!

PR #93, opened 2025-04-29
[DESCW-2991] Add DatabaseConfig::getConnectionArray()

mhaswell-bcgov:
Update Database class and migrations-db script to use above
Fix typo in NaadSocketClientTest function

### DIFF
--- a/migrations-db.php
+++ b/migrations-db.php
@@ -2,11 +2,8 @@
 
 // See https://www.doctrine-project.org/projects/doctrine-migrations/en/3.8/reference/configuration.html#simple
 
-return [
-    'user'     => 'root',
-    'password' => $_ENV['MARIADB_ROOT_PASSWORD'],
-    'host'     => $_ENV['MARIADB_SERVICE_HOST'],
-    'port'     => $_ENV['MARIADB_SERVICE_PORT'],
-    'dbname'   => $_ENV['MARIADB_DATABASE'],
-    'driver'   => 'pdo_mysql',
-];
+use Bcgov\NaadConnector\Config\DatabaseConfig;
+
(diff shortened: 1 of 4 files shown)


ShawnTurple (changes requested):
Looks good, although we should update the version, and assuming we are going to use releases, we don't need to update the changelog, and that information will be obvious in the release.

mhaswell-bcgov:
> Looks good, although we should update the version, and assuming we are going to use releases, we don't need to update the changelog, and that information will be obvious in the release.

@ShawnTurple I removed the version and added a license to composer.json per `composer validate` recommendations:
```
# General warnings
- No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
- The version field is present, it is recommended to leave it out if the package is published on Packagist.
```

composer.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "bcgov/naad-connector",
-  "version": "1.0.0",
```
ShawnTurple, line 3: why did you remove this?

mhaswell-bcgov: It serves no purpose in this project since we're not installing this project anywhere via composer. When you run `composer validate` it actually flags having the `version` field at all as a warning:
```
- The version field is present, it is recommended to leave it out if the package is published on Packagist.
```

ShawnTurple: ok, so then how are we keeping track of versions, agree that in this case it is not used, but when we are developing things, how is the version going to be tracked, if you say simply tagging, then i would say we have a few processes to work out.

ShawnTurple (approved):
i'm ok approving this, but i suspect there are still some process things we need to figure out with versions

PR #94, opened 2025-04-29
DESCW-2924 concentrates test .env to a single file.

robrien-bcgov:


tests/unit/bootstrap.php
```diff
@@ -0,0 +1,15 @@
...
+ * - Loads the Composer autoloader to include dependencies.
+ * - Clears all environment variables from the $_ENV superglobal to ensure
+ *   a clean testing environment.
+ *
+ */
+<?php
+require_once __DIR__ . '/../../vendor/autoload.php';
+
+foreach (array_keys($_ENV) as $key) {
+    unset($_ENV[$key]);
```
mhaswell-bcgov, line 14: This doesn't seem to work the way I would expect. In `.devcontainer/devcontainer.env` I set `LOG_RETENTION_DAYS=6` and then in `tests/unit/.env.test` I commented out `LOG_RETENTION_DAYS` so I would expect the tests to not have a value for LOG_RETENTION_DAYS but I'm still getting: 
```
Logger Config
 ✘ Magic getter
   ┐
   ├ Failed asserting that 6 matches expected 7.
   │
   │ /workspaces/naad-connector/tests/unit/LoggerConfigTest.php:69
```
So I'm still getting the `6` from the devcontainer env but I would expect it to get wiped out here.

mhaswell-bcgov (changes requested):

tests/unit/bootstrap.php
robrien-bcgov: interesting. I'll have to see if it's stored in the $_SERVER superglobal or something like that

robrien-bcgov: Ok. This should clear out those stale environment variables (just the naad-connector ones).

mhaswell-bcgov: This seems to have the same problem:
```
...
Unsetting stale environment variable: ALERTS_TO_KEEP
Unsetting stale environment variable: LOG_PATH
Unsetting stale environment variable: LOG_RETENTION_DAYS
...
Logger Config
 ✘ Magic getter
   ├ Failed asserting that 6 matches expected 7.
   │ /workspaces/naad-connector/tests/unit/LoggerConfigTest.php:69
...
```
It doesn't seem like the bootstrap script runs at the right time to actually unset the env variables for the tests. I'd be ok with just the .env.test ensuring the env variables we know about are overridden and removing the bootstrap file.

robrien-bcgov: That's strange. It worked on my machine, and I did the same thing.
image
I even ran the client to see if that would cause problems.

mhaswell-bcgov: The difference is I removed `LOG_RETENTION_DAYS` from .env.test but it looks like yours is unchanged.

mhaswell-bcgov (changes requested):
From a quick look at the env variables we're using in the config classes, I think we're just missing `ALERTS_TO_KEEP` in `.env.test` so if we add that (set to `100`) I think we're good.

tests/unit/bootstrap.php
robrien-bcgov: at least removing the <env> from the test config xml file seems to force the local .env.test to apply. I would guess that the value in .devcontainer's .env is refreshed outside of the bootstrap lifecycle - perhaps in server superglobal or system environment (inside the running container?), but It's no big deal as long as this works once deployed.

I noticed the workflow seems to have responded differently to having all the naad-connector $_ENV values unset - It couldn't find any of them when run inside whatever container the workflow uses while testing.

mhaswell-bcgov (approved):

PR #98, opened 2025-05-07
DESCW-2997 Replace mariadb root user with a new service user for auth

kshapka-bcgov:
- Adds a new job which creates a service user for mariadb authentication
  - Also ads postsync waves for Argo CD so that this job _should_ run before the migrations
  - Also slightly increases the delay between retries on migrations to accommodate deployments run without Argo CD
- Replaces mariadb root user authentication with new service user
- Modifies tests to include the new user instead of root

Notes: Requires the addition of the following variables to your .env for deployment (see sample-env):
`MARIADB_SERVICE_USER=naad_user`
`MARIADB_SERVICE_PASSWORD=password`


sample-env
mhaswell-bcgov, line 1: Do we still use env variables from here? I've only been using the devcontainer to run things locally and I'm not sure how to make this work there. We'd need to update `.devcontainer/devcontainer.env` but we'd still need a way to run the user create command (like this: https://github.com/bcgov/naad-connector/blob/feature/DESCW-2997/.devcontainer/devcontainer.json#L24) since the Job won't run in the devcontainer.

mhaswell-bcgov (changes requested):
This is cool. I have some ideas on how to restrict the permissions of the user even more: https://github.com/bcgov/naad-connector/compare/feature/DESCW-2997...restrict-user-permissions.
My idea is that the migrations run under the root user but the application runs under the less-privileged user that can only SELECT, INSERT, UPDATE, and DELETE but can't do anything to the structure of the database/tables.

sample-env
kshapka-bcgov: We can't use more limited permissions because then the migration job will fail. That would require reworking on its own.

mhaswell-bcgov:
> > This is cool. I have some ideas on how to restrict the permissions of the user even more: [feature/DESCW-2997...restrict-user-permissions](https://github.com/bcgov/naad-connector/compare/feature/DESCW-2997...restrict-user-permissions). My idea is that the migrations run under the root user but the application runs under the less-privileged user that can only SELECT, INSERT, UPDATE, and DELETE but can't do anything to the structure of the database/tables.
> 
> We can't limit the permissions further because the migrations currently use the same user. Changing that would require work of its own.

I have this working in the branch I linked: https://github.com/bcgov/naad-connector/tree/restrict-user-permissions. The migrations run under `root` but the application runs under `naad_user`.

.devcontainer/devcontainer.env
```diff
@@ -1,5 +1,6 @@
 # Database environment vars
-MARIADB_ROOT_PASSWORD=rootpassword
+MARIADB_SERVICE_USER=naad_user
+MARIADB_SERVICE_PASSWORD=password
```
mhaswell-bcgov, line 3:
```suggestion
MARIADB_SERVICE_USER=root
MARIADB_SERVICE_PASSWORD=rootpassword
```
I think for now it'll have to run everything under `root` in the devcontainer because there's nothing that creates the user when the devcontainer runs. We can add this in a future ticket.

mhaswell-bcgov (commented):
Looks like the dev instance's application pod is failing due to database user access being denied. I don't know if it's a just configuration issue or if something needs to change in the code.

.devcontainer/devcontainer.env
kshapka-bcgov: Yeah, I think we can run a command to add the user in the devcontainer json, but I was having trouble getting it working. Something like:
`"postCreateCommand": [the same command from the job]`

mhaswell-bcgov: Yeah I played with it a bit to see if I could get that working but I couldn't. I think it's ok for now.

mhaswell-bcgov (approved):

PR #99, opened 2025-05-13
DESCW-2998 continue fetching if palmorex returns 500 + test

robrien-bcgov:
This pull request removes the custom `AlertFetchFailureException` class and updates the error handling logic across the `NaadConnector` codebase to directly use standard exceptions, such as `RequestException`. Additionally, it enhances the handling of missed alerts during heartbeat processing and introduces new test cases to ensure robust error handling.

### Removal of `AlertFetchFailureException`:

* Deleted the `AlertFetchFailureException` class from `src/Exception/AlertFetchFailureException.php` and removed its usage across the codebase. Error handling now directly uses `RequestException` or other standard exceptions. [[1]](diffhunk://#diff-5d1eaa9f14b6edcd82b757a8151f2bd75bd4b5822ff6e87ad0e191e03c1c40e7L1-L33) [[2]](diffhunk://#diff-5055adc2b6a425aaa3f513b9ac294ba7420220d9a79e07893561fe964438f61aL4) [[3]](diffhunk://#diff-5055adc2b6a425aaa3f513b9ac294ba7420220d9a79e07893561fe964438f61aL73-R72)

### Updates to Error Handling:

* Updated the `fetchAlert` method in `NaadRepositoryClient` to throw `RequestException` directly instead of wrapping it in a custom exception. [[1]](diffhunk://#diff-5055adc2b6a425aaa3f513b9ac294ba7420220d9a79e07893561fe964438f61aL56-R55) [[2]](diffhunk://#diff-5055adc2b6a425aaa3f513b9ac294ba7420220d9a79e07893561fe964438f61aL73-R72)
* Enhanced the `handleHeartbeat` method in `NaadSocketClient` to:
  - Log errors and continue processing missed alerts for exceptions with HTTP status codes >= 500.
  - Re-throw exceptions with codes < 500 to ensure critical issues are not ignored.

### Test Enhancements:

* Removed the `AlertFetchFailureExceptionTest` class, as the custom exception is no longer in use.
* Updated `NaadRepositoryClientTest` to reflect the removal of `AlertFetchFailureException`:
  - Adjusted test cases to expect `RequestException` instead. [[1]](diffhunk://#diff-9999fb8033271eae2aa3381510b8587f016d0967bdfce15c9a802484f16584dcL5) [[2]](diffhunk://#diff-9999fb8033271eae2aa3381510b8587f016d0967bdfce15c9a802484f16584dcL32) [[3]](diffhunk://#diff-9999fb8033271eae2aa3381510b8587f016d0967bdfce15c9a802484f16584dcL88-R86) [[4]](diffhunk://#diff-9999fb8033271eae2aa3381510b8587f016d0967bdfce15c9a802484f16584dcL102-R100) [[5]](diffhunk://#diff-9999fb8033271eae2aa3381510b8587f016d0967bdfce15c9a802484f16584dcL132-R130)
* Added new test cases in `NaadSocketClientTest` to verify:
  - Proper error logging for exceptions with HTTP codes >= 500 during heartbeat processing.
  - Correct exception re-throwing behavior for exceptions with codes < 500.

### Miscellaneous:

* Added a new XML file, `heartbeat-missed-alert.xml`, to simulate missed alerts during heartbeat tests.


tests/unit/NaadSocketClientTest.php
```diff
@@ -466,4 +466,128 @@ public function testHandleResponseDestinationClientException()
...
+
+        $client = new NaadSocketClient(
+            $destinationClient,
+            $logger,
+            $database,
+            $repositoryClient
+        );
+
+        // Prepare a missed alert to trigger fetchAlert
+        $reflection = new \ReflectionClass($client);
```
mhaswell-bcgov, line 493: I don't think we need to use Reflection here. I'd rather avoid it if possible. We should be able to get the same results by calling the public function `handleResponse()`.

tests/unit/NaadSocketClientTest.php
```diff
@@ -466,4 +466,128 @@ public function testHandleResponseDestinationClientException()
...
+        $method = $reflection->getMethod('handleHeartbeat');
+        $method->setAccessible(true);
+
+        // Simulate findMissedAlerts to return one alert
+        $findMissedAlerts = $reflection->getMethod('findMissedAlerts');
+        $findMissedAlerts->setAccessible(true);
```
mhaswell-bcgov, line 499:
```suggestion
```
This doesn't seem necessary.

src/NaadSocketClient.php
```diff
@@ -155,9 +155,26 @@ protected function handleHeartbeat(SimpleXMLElement $xml)
...
-                $rawXml = $this->repositoryClient->fetchAlert(
-                    $alert['id'], $alert['sent']
-                );
+                try {
+                    $rawXml = $this->repositoryClient->fetchAlert(
+                        $alert['id'], $alert['sent']
+                    );
+                } catch (Exception $e) {
+                    // if the exception is a 500 error, log it and continue
+                    if ($e->getCode() === 500) {
```
mhaswell-bcgov, line 164:
```suggestion
                    if ($e->getCode() >= 500) {
```
All 500+ level status codes should be handled this way.

mhaswell-bcgov (dismissed):
I tested this with the pelmorex site by giving it a bad url (causing a 404) but when it gets to the check, `$e->getCode()` is not the status code (it's `0`), so this logic I believe is right but the `AlertFetchFailureException` we're throwing is hiding the status code. I actually don't think we need that exception at all and could remove it from the repo since it was only added to satisfy a sonar issue. We should have fetchAlert() not catch anything which will cause it to throw a `GuzzleHttp\Exception\RequestException` whose `getCode()` _will_ return the status code.

Changes requested:

1. Remove the try/catch from `NaadRepositoryClient::fetchAlert()`, allowing it to just throw the `GuzzleHttp\Exception\RequestException`.
2. Remove `AlertFetchFailureException` and its unit tests.
3. The comments made on specific lines (checking for >500 status code and reflection in unit tests).

tests/unit/NaadSocketClientTest.php
robrien-bcgov: I think the error code wasn't passed to that AlertFetchFailureException instance in the first place.

It's cleaner without the class though.

mhaswell-bcgov (changes requested):
@robrien-bcgov Thanks for making the changes, looks like it's working now. I also made some changes to the unit tests, see the commit message for the reasoning.
Your two most recent commits although I see that they're addressing sonar issues I'd rather do them in another ticket/PR since it's complicating this review to have make sure those are now working as expected too. Can you revert those changes?

robrien-bcgov:
@mhaswell-bcgov  changes reverted for the two commits behind yours

mhaswell-bcgov (approved):
Looks good, thanks!

